PR #29, opened 2019-04-02
Fix most of the issues reported by Go Report

Crevil:
GoReport have found a great deal of issues in our code base. This PR fixes most of them.

Only thing I haven't looked into is the documentation of all exported methods. I think we should do this over time, but it's a bit excessive right now.


kaspernissen:
\lgtm

PR #43, opened 2019-04-11
Rollback command

Crevil:
This PR implements a `rollback` command. It will release the artifact that was running in an environment before the current one.

```
$ hamctl rollback --service a --env dev
[✓] Rollback of artifact 'master-a-2' initiated
    Release of 'master-a-1' to 'dev'
```

The result is seen here: https://github.com/lunarway/release-manager-test-config-repo/commit/db32c5ac0c17b2b495791df77b17b6679c1ea1f4

### Other changes
I made the `make` targets for `artifact` build the CLI once instead of on every step. Increases the speed of the targets by much!

I also ensured that function `git.LocateRelease()` only identifies release commits, ie. `release <some-artifact-id>`. The current implementation just matches the artifact ID, which will clash with the `rollback <artfifact-current> to <artifact-previous>` messages of this command.

### DIFF
--- a/internal/flow/rollback.go
+++ b/internal/flow/rollback.go
@@ -1,0 +1,102 @@
+package flow
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path"
+
+	"github.com/lunarway/release-manager/internal/git"
+	"github.com/lunarway/release-manager/internal/log"
+	"github.com/otiai10/copy"
+	"github.com/pkg/errors"
+)
+
+type RollbackResult struct {
+	Previous string
+	New      string
+}
+
+func Rollback(ctx context.Context, configRepoURL, artifactFileName, service, env, committerName, committerEmail, sshPrivateKeyPath string) (RollbackResult, error) {
+	r, err := git.Clone(ctx, configRepoURL, sourceConfigRepoPath, sshPrivateKeyPath)
+	if err != nil {
+		return RollbackResult{}, errors.WithMessagef(err, "clone '%s' into '%s'", configRepoURL, sourceConfigRepoPath)
+	}
+
+	// locate current release
+	currentHash, err := git.LocateServiceRelease(r, env, service)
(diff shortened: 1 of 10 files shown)


internal/flow/rollback.go
```diff
@@ -0,0 +1,102 @@
...
+		return RollbackResult{}, errors.WithMessage(err, fmt.Sprintf("copy artifact spec from '%s' to '%s'", artifactSourcePath, artifactDestinationPath))
+	}
+
+	authorName := newSpec.Application.AuthorName
+	authorEmail := newSpec.Application.AuthorEmail
+	releaseMessage := git.RollbackCommitMessage(env, service, currentSpec.ID, newSpec.ID)
+	err = git.Commit(ctx, destinationRepo, releasePath(".", service, env), authorName, authorEmail, committerName, committerEmail, releaseMessage, sshPrivateKeyPath)
+	if err != nil {
+		return RollbackResult{}, errors.WithMessage(err, fmt.Sprintf("commit changes from path '%s'", destinationPath))
+	}
```
kaspernissen: We should notify the release channels here. 


```
	err = notifyRelease(slack.ReleaseOptions{
		SlackToken:    slackToken,
		Service:       service,
		Environment:   env,
		ArtifactID:    sourceSpec.ID,
		CommitAuthor:  sourceSpec.Application.AuthorName,
		CommitMessage: sourceSpec.Application.Message,
		CommitSHA:     sourceSpec.Application.SHA,
		CommitLink:    sourceSpec.Application.URL,
		Releaser:      committerName,
	})
	if err != nil {
		log.Errorf("flow: Promote: error notifying release: %v", err)
	}
```

Crevil: You are absolutely right 👍 I'll add that.

kaspernissen: I just added it!

PR #49, opened 2019-04-17
add usermappings and checks to server and artificat

kaspernissen:
This PR adds a user mappings check to enable mapping of non Lunar Way emails to Lunar Way emails, that is needed in terms of notifying users via Slack


cmd/artifact/command/init.go
```diff
@@ -14,11 +17,47 @@ import (
...
+				users = strings.Split(userMappingString, ",")
+			}
+			m := make(map[string]string)
+			for _, u := range users {
+				s := strings.Split(u, "=")
+				m[s[0]] = s[1]
+			}
+			options.UserMappings = m
+
+			fmt.Printf("EMAIL: %s\n", s.Application.AuthorEmail)
```
Crevil: This should not be here, I guess?

cmd/server/command/root.go
```diff
@@ -34,6 +49,7 @@ func NewCommand() (*cobra.Command, error) {
 	command.PersistentFlags().StringVar(&options.GrafanaDevUrl, "grafana-dev-url", os.Getenv("GRAFANA_DEV_URL"), "grafana dev url")
 	command.PersistentFlags().StringVar(&options.GrafanaStagingUrl, "grafana-staging-url", os.Getenv("GRAFANA_STAGING_URL"), "grafana staging url")
 	command.PersistentFlags().StringVar(&options.GrafanaProdUrl, "grafana-prod-url", os.Getenv("GRAFANA_PROD_URL"), "grafana prod url")
+	command.PersistentFlags().StringSliceVar(&users, "user-mappings", []string{}, "user mappings between github and slack, key-value pair: <github>=<slack>")
```
Crevil: Maybe mention that the mapping is of emails. Further more, it can be used for what ever committer info, not just GitHub, right?

cmd/artifact/command/init.go
kaspernissen: No, good catch!

cmd/server/command/root.go
kaspernissen: That's true. Good point!

PR #52, opened 2019-04-23
add slack notification if auto-release policy fails

kaspernissen:
This PR adds slack notifications to the git author in case of auto-release failures.

### DIFF
--- a/cmd/server/http/http.go
+++ b/cmd/server/http/http.go
@@ -568,6 +568,7 @@ func notifyPolicyFailure(email, title, errorMessage, token string, userMappings 
 	client, err := slack.NewClient(token)
 	if err != nil {
 		log.Errorf("error initializing Slack Client in notifyPolicyFailure: %v", err)
+		return
 	}
 
 	if !flow.IsLunarWayEmail(email) {
--- a/cmd/server/http/http.go
+++ b/cmd/server/http/http.go
@@ -575,6 +575,7 @@ func notifyPolicyFailure(email, title, errorMessage, token string, userMappings 
 		lwEmail, ok := userMappings[email]
 		if !ok {
 			log.Errorf("%s is not a Lunar Way email and no mapping exist", email)
+			return
 		}
 		email = lwEmail
 	}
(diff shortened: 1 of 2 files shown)


cmd/server/http/http.go
```diff
@@ -561,3 +563,27 @@ func getGrafanaVarForEnv(env, dev, staging, prod string) string {
 	}
 	return ""
 }
+
+func notifyPolicyFailure(email, title, errorMessage, token string, userMappings map[string]string) {
+	client, err := slack.NewClient(token)
+	if err != nil {
+		log.Errorf("error initializing Slack Client in notifyPolicyFailure: %v", err)
```
Crevil: You need to return after this line to avoid continuing.
```suggestion
		log.Errorf("error initializing Slack Client in notifyPolicyFailure: %v", err)
		return
```

cmd/server/http/http.go
```diff
@@ -561,3 +563,27 @@ func getGrafanaVarForEnv(env, dev, staging, prod string) string {
...
+	}
+
+	if !flow.IsLunarWayEmail(email) {
+		//check UserMappings
+		lwEmail, ok := userMappings[email]
+		if !ok {
+			log.Errorf("%s is not a Lunar Way email and no mapping exist", email)
```
Crevil: You need to return after this line to avoid continuing.

```suggestion
			log.Errorf("%s is not a Lunar Way email and no mapping exist", email)
			return
```

cmd/server/http/http.go
```diff
@@ -561,3 +563,27 @@ func getGrafanaVarForEnv(env, dev, staging, prod string) string {
...
+		}
+		email = lwEmail
+	}
+	slackId, err := client.GetSlackIdByEmail(email)
+	if err != nil {
+		log.Errorf("error obtaining slackId in notifyPolicyFailure: %v", err)
```
Crevil: You need to return after this line to avoid continuing.

```suggestion
		log.Errorf("error obtaining slackId in notifyPolicyFailure: %v", err)
		return
```

kaspernissen:
Yes, of course! Thanks!

PR #55, opened 2019-04-23
Implement service structs

Crevil:
Change `flow`, `policy` and `grafana` functions to methods. This reduces the
number of arguments and the general overhead of non-important arguments
on each function call.

I'm unsure how to handle the user email mapping in the `notifyPolicyFailure` function. It seems a bit weird that you need to lookup the email up front instead of just letting the `slack.Client` handle this.
What do you think of letting the `Client` handle the mapping. The general API of the `Client` struct is based on Slack user IDs. We could change this to emails instead of do the lookup first thing?


kaspernissen:
Yes, i thought about this as well. It makes sense to move the email lookup into the slack client and handle the user mapping. It's the only place it is used.

PR #60, opened 2019-04-25
make sure to init log in artifact

kaspernissen:
This PR inits the logging service in artifact to fix the problem Nicolai experienced yesterday.


kaspernissen:
I will just merge it to get it out.

PR #61, opened 2019-04-25
add option to post custom messages in case of failure

kaspernissen:
This PR adds an option for adding an error-message to the `artifact failure` command to post custom error messages

### DIFF
--- a/cmd/artifact/command/failure.go
+++ b/cmd/artifact/command/failure.go
@@ -14,12 +14,6 @@ func failureCommand(options *Options) *cobra.Command {
 	command := &cobra.Command{
 		Use:   "failure",
 		Short: "report failure in the pipeline",
-		PreRunE: func(cmd *cobra.Command, args []string) error {
-			if errorMessage == "" {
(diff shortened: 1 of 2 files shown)


cmd/artifact/command/failure.go
```diff
@@ -10,9 +10,16 @@ import (
 )
 
 func failureCommand(options *Options) *cobra.Command {
+	var errorMessage string
 	command := &cobra.Command{
 		Use:   "failure",
 		Short: "report failure in the pipeline",
+		PreRunE: func(cmd *cobra.Command, args []string) error {
```
Crevil: This could be `PreRun` instead to avoid the `error` return value.

cmd/artifact/command/failure.go
```diff
@@ -55,5 +62,6 @@ func failureCommand(options *Options) *cobra.Command {
 			return nil
 		},
 	}
+	command.Flags().StringVar(&errorMessage, "error-message", "", "error message to send to slack")
```
Crevil: You can add the default value here instead of in `PreRunE`?

PR #64, opened 2019-04-29
add if statement to ensure looking in the correct folder

kaspernissen:
This PR seems to fix the issue with promote. We were looking in the wrong folder when locating the release-spec. We need to look at the previous env namespace folder in the given env.


kaspernissen:
I will just merge, as @Crevil has approved IRL.

PR #69, opened 2019-05-06
ensure better environment detection

kaspernissen:


cmd/server/http/http.go
```diff
@@ -265,12 +266,12 @@ func daemonWebhook(flowSvc *flow.Service) http.HandlerFunc {
 
 		err = flowSvc.NotifyCommitter(context.Background(), &podNotify)
 		if err != nil {
-			logger.Errorf("http: daemon webhook failed: pod '%s' namespace '%s': notify committer: %v", podNotify.Name, podNotify.Namespace, err)
+			logger.Errorf("http: daemon webhook failed: pod '%s' namespace '%s' environment: %s notify committer: %v", podNotify.Name, podNotify.Namespace, podNotify.Environment, err)
```
Crevil: Can we add `'` quotes around the environment string like you do below? This will make it super clear if the field for some reason is empty.

internal/flow/notify.go
```diff
@@ -44,15 +44,11 @@ func (s *Service) NotifyCommitter(ctx context.Context, event *http.PodNotifyRequ
...
 	}
-	env := matches[1]
+	// Use environment received from release-daemon
+	env := event.Environment
 	service := matches[2]
 	namespace := event.Namespace
-	// handle cases where the namespace would default to the environment
-	// this ensures that we can locate artifact specifications on deployments in
-	// these namespaces
-	if isEnvironmentNamespace(event.Namespace) {
```
Crevil: You can delete the function `isEnvironmentNamespace` as well now, as this was its only usage.

PR #76, opened 2019-05-13
send reference to userMappings instead, and dereference in start instead

kaspernissen:
This PR fixes the problem of userMappings not being parsed down the chain to the Slack client.

### DIFF
--- a/internal/slack/slack.go
+++ b/internal/slack/slack.go
@@ -23,7 +23,7 @@ var (
 )
 
 func NewClient(token string, emailMappings map[string]string) (*Client, error) {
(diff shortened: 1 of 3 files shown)


internal/slack/slack.go
```diff
@@ -23,6 +23,7 @@ var (
 )
 
 func NewClient(token string, emailMappings map[string]string) (*Client, error) {
+	log.Infof("slack; new client; initialized with emailMappings: %+v", emailMappings)
```
Crevil:
```suggestion
	log.Infof("slack: new client: initialized with emailMappings: %+v", emailMappings)
```

PR #79, opened 2019-05-29
Add a message that instructs the user to release manually with hamctl

kaspernissen:
This PR outputs a more clear message when an autorelease policy fails. We instruct the user to to a release manually


Crevil:
I am!!!

PR #93, opened 2019-06-29
Use native git for performance

Crevil:
This change replaces [`src-d/go-git`](https://github.com/src-d/go-git) with the native `git` CLI for most repository interactions.

This greatly reduces request response times and memory allocations.

I also made the deletion of master repository copies asynchronous to avoid letting clients wait for the clean up to be done.

With this change the `Locate*` functions in package `git` are updated to accept a trailing newline in commit messages, as the `git commit` command enforces this where as `go-git` did not.

Lastly I modified the retry-functionality of all `flow` methods to sync the master repo if we detect that we are out of sync during a commit.
This should mitigate the errors we have seen when GitHub webhooks does not arrive to update the master repo.

### DIFF
--- a/internal/flow/describe.go
+++ b/internal/flow/describe.go
@@ -85,7 +85,7 @@ func (s *Service) DescribeArtifact(ctx context.Context, service string, n int) (
 	var artifacts []artifact.Spec
 	log.Debugf("flow/describe: hashes %+v", hashes)
 	for _, hash := range hashes {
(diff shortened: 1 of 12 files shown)


Crevil:
@kaspernissen please use this PR's description as the commit message if you merge this to keep a clear log of the changes it contains. The commit history is not very interesting. :)

kaspernissen:
Looks really good! I'm testing this in our env today, and merge tomorrow if all goes well 👍

PR #111, opened 2019-10-18
Fx bug where release logs does not contain fields

Crevil:


Crevil:
I'll merge right away as our release tracking is broken by this.

PR #120, opened 2019-12-03
Add request IDs to traces and logs

Crevil:
Request IDs are generated by hamctl and passed in an HTTP header to the server.
The server will tag request logs and traces with the request ID.

If no request ID is set in the incoming request the server generates one.

### DIFF
--- a/cmd/hamctl/main.go
+++ b/cmd/hamctl/main.go
@@ -28,7 +28,6 @@ func main() {
 	c.AddCommand(versionCmd)
 	err = c.Execute()
 	if err != nil {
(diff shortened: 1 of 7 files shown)


cmd/hamctl/main.go
```diff
@@ -28,7 +28,6 @@ func main() {
 	c.AddCommand(versionCmd)
 	err = c.Execute()
 	if err != nil {
-		fmt.Printf("Error: %v\n", err)
```
Crevil: Cobra already logs errors from `Execute` so this is removed to avoid printing the error and reference twice.

PR #121, opened 2019-12-05
Support prerelease tags

Crevil:
This change deprecates the mutable docker tags `latest`, `vX` and `vX.Y`. We do not use
these any ways and Goreleaser don't have conditionally selecting the docer image
templates.

This unlocks the posibility of pushing `v1.2.3-alpha1` tags that gets build and
released to GitHub as prerelease and as Docker images with the complete tag
name.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -48,20 +48,14 @@ dockers:
     binaries:
     - server
     image_templates:
-    - "quay.io/lunarway/release-manager:latest"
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -82,6 +76,7 @@ snapshot:
 
 release:
   name_template: "v{{.Version}}"
+  prerelease: auto
```
Crevil: This marks GitHub releases as pre-release if a pre-suffix is detected in the tag.

PR #124, opened 2019-12-05
Upgrade Go and Alpine to latest

Crevil:
Awaits the Go 1.13.5 Docker image is available on [Docker hub](https://hub.docker.com/_/golang/).


Crevil:
The docker image is now available.

PR #140, opened 2020-03-15
Make muteable endpoints asyncronous

Crevil:
This is a rather large PR, but the changes to existing logic is kept limited, so brace yourself. 🙈 

In order to provide a better developer experience on large config repositories (>10000 commits) where release and other mutating commands are getting slow, ie. 10s of seconds, we are going to make it fully asynchronous.

This means that we only do some initial input validation before publishing an event to RabbitMQ and returning a response to the client. The server will then consume these events one by one.

This will reduce the response time of mutating commands and further ensure that their operation is completed as soon as possible instead of letting the client in charge for retrying.
It will also reduce the number of conflicts when multiple clients are trying to write to the cluster config repository. Before this, clients had to inspect the response in this case and retry the attempts, now it will be handled by only handling a single event at a time.

### AMQP publisher and consumer
Package `amqp` implements an AMQP `Worker` capable of handling connection failures and other fatal errors on the AMQP connection.
It exposes two methods providing the general functionality of the AMQP protocol: `StartConsumer` and `Publish`.

The former is a blocking method that starts consuming messages on the configured queue with provided handlers for the different event types.
The latter publishes a message on the configured exchange.

### Flows

Release, promote and rollback flows are updated to only do required input validation, ie. check for available artifacts to the environments. They will then publish a message on AMQP that is handled by the `Exec*` variants of the methods that performs the actual manipulation of the config repository.

## Open TODOs

There are a number of improvements that could be made that I've only noted as TODOs right now. They'll be fixed in upcoming PRs to keep things simple and get this into master.

### Non-async flow
A flag to replace the publisher/consumer with an in-memory queue to ease local development and deployment.

### Nack events with dead letter queue
Events are not nacked/acked in case of event handler errors. With a low prefetch (default `1`) this means that event handling is blocked after the first failed event. 

To fix this we can either just drop the events (nack and no requeue) and inform the user of what happened. Another option is to add a dead letter queue and push events over there. This does however require a mechanism for persisting them from there and further redelivering/dropping them.

I'm most in favour on the former one. Errors could be like invalid configuration (no config for environment) etc and these bugs should be handled in the sync validation step instead.

### Cancel consumer before closing it
The `Close` method of the `consumer` type in package `amqp` closes its AMQP channel but does not cancel it before hand. This can potentially lead to dropping inflight events as the channel is closed before all events are consumed.

We can handle this by adding a wait group or something like that guarding the `Start` method that is the consumer. With this we can cancel the channel, wait for the consumer to stop (no more messages) and then close the channel.

### Publish confirms
AMQP provides a mechanism for guaranteeing published messages to have arrived on consumer queues. This is not handled for now and can potentially drop events if the remote AMQP server fails in forwarding the message to queues.

### Environment up to date messages
Going async means that the git status identifying that we have nothing new to release is handled in the async execution path. There is no message to the releaser for now except a log message so eventually we want to push this message out in Slack.
A even better solution would be to validate the released artifact ID against the current one in the target environment and thus provide this message in the synchronous flow.


internal/amqp/consumer.go
```diff
@@ -0,0 +1,131 @@
...
+		if err != nil {
+			logger.With("res", map[string]interface{}{
+				"status":       "failed",
+				"responseTime": duration,
+				"error":        fmt.Sprintf("%+v", err),
+			}).Errorf("[consumer] [FAILED] Failed to handle message: nacking and requeing: %v", err)
+			// TODO: remove comments to allow for redelivery. This will put events
+			// into the unacknowledged state
+
+			// err := msg.Nack(false, true) if err != nil {
```
kaspernissen: Should the comments be removed? :)

Crevil:
There is still a couple of todos in here. The handling of no changes (environment up to date) in particular.
I got an idea on how to do it in the validation step, but I can wait with that to another PR just to keep this a bit simpler.

internal/amqp/consumer.go
Crevil: Hmm. Yes. The unacked mechanism is critical when having preferch of one, as an error will block releases. But uncommenting can result in looping failed events leading to high loads.

The right thing to do is to add a dead letter queue on the messages. I think we can do that rather simple, but in another PR

kaspernissen: Sounds good 👍

Crevil:
@emilingerslev I jut pinged you for a review. Note the `Open TODOs` in the description before diving into the code :)

Crevil:
@tbflw could I maybe get your review of package `amqp` as well?

We talked about it last week regarding the implementation, but as you also know our internal amqp library, maybe you could chime in here? I'm hoping that we can change the internal impementation to use this instead as I see it very much simpler.

internal/flow/release.go
```diff
@@ -24,8 +24,65 @@ import (
 func (s *Service) ReleaseBranch(ctx context.Context, actor Actor, environment, service, branch string) (string, error) {
 	span, ctx := s.Tracer.FromCtx(ctx, "flow.ReleaseBranch")
 	defer span.Finish()
-	var result string
+	sourceConfigRepoPath, close, err := git.TempDirAsync(ctx, s.Tracer, "k8s-config-release-branch")
```
emilingerslev: Why so much new code? Is it moved from somewhere?

internal/amqp/publisher.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (p *rawPublisher) Close() error {
+	err := p.channel.Close()
+	if err != nil {
+		return errors.WithMessage(err, "close amqp channel")
+	}
+	return nil
+}
+
+type loggingPublisher struct {
```
Crevil: Move out of the package as it is not relevant for the amqp package.

cmd/server/command/start.go
```diff
@@ -149,6 +173,78 @@ func NewStart(grafanaOpts *grafanaOptions, slackAuthToken *string, githubAPIToke
...
+					Port:        amqpOptions.Port,
+				},
+				MaxReconnectionAttempts: amqpOptions.MaxReconnectionAttempts,
+				ReconnectionTimeout:     amqpOptions.ReconnectionTimeout,
+				Exchange:                amqpOptions.Exchange,
+				Queue:                   amqpOptions.Queue,
+				RoutingKey:              "#",
+				Prefetch:                amqpOptions.Prefetch,
+				Logger:                  log.With("system", "amqp"),
+				Handlers: map[string]func(d []byte) error{
```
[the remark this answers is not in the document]
Crevil: We will log an error and drop the received event.

internal/amqp/amqp.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+	worker.config.Logger = c.Logger.WithFields(
+		"host", c.Connection.Host,
+		"user", c.Connection.User,
+		"port", c.Connection.Port,
+		"virtualHost", c.Connection.VirtualHost,
+		"exchange", c.Exchange,
+		"queue", c.Queue,
+		"prefetch", c.Prefetch,
+		"maxReconnectionAttempts", c.MaxReconnectionAttempts,
```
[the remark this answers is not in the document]
Crevil: No reason really. After talking about this we decided to remove this functionality and instead keep retrying in case of lost connections.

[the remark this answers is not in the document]
emilingerslev: 👍

internal/amqp/amqp.go
```diff
@@ -0,0 +1,286 @@
...
+	err = s.currentPublisher.Publish(ctx, event.Type(), uuid.String(), event.Body())
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (s *Worker) connect() error {
+	c := s.config
+	c.Logger.Infof("Connecting to: %s", c.Connection.String())
```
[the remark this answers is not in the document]
Crevil: Moved all the config fields to a dedicated logger in `connect` to make this simpler and avoid having config on all publish/consume log messages.

internal/flow/release.go
emilingerslev: With indentions removed it made a lot more sense. Thanks for explaining 👍

internal/amqp/amqp.go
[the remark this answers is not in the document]
emilingerslev: Awesome 👍

internal/amqp/publisher.go
```diff
@@ -0,0 +1,100 @@
...
+	exchangeName string
+	routingKey   string
+}
+
+func newPublisher(amqpConn *amqp.Connection, exchangeName string) (publisher, error) {
+	channel, err := amqpConn.Channel()
+	if err != nil {
+		return nil, errors.WithMessage(err, "get channel")
+	}
+	err = declareExchange(channel, exchangeName)
```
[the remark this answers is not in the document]
Crevil: Done 👍

internal/flow/release.go
Crevil: We found out that showing the diff without whitespace changes fixed that 🎉

Crevil:
I've updated the things you've mentioned @emilingerslev  in the review and will leave the last parts like the general decorator pattern on the publisher for another PR.

emilingerslev:
Sounds good, will try to look it over 👍

internal/amqp/publisher.go
emilingerslev: You'll also wait with moving this one out?

Crevil:
Yeah I thought about it, but would rather move it around along with the message id generator, tracing etc so I have more use cases for designing the decorator API.

PR #141, opened 2020-03-16
Support for running fluxd, manager and daemon locally

emilingerslev:


Makefile
```diff
@@ -173,3 +173,49 @@ server-profile-cpu:
 jaeger:
 	open http://localhost:16686
 	docker run --rm -p 5775:5775/udp -p 6831:6831/udp -p 6832:6832/udp -p 5778:5778 -p 16686:16686 -p 14268:14268 -p 9411:9411 -e COLLECTOR_ZIPKIN_HTTP_PORT=9411 jaegertracing/all-in-one:1.7
+
+e2e-setup:
+	make e2e-setup-git
+	make e2e-setup-kind
+	make e2e-setup-fluxd
```
Crevil: Any reason for not just running these as make pre-targets?

```suggestion
e2e-setup: e2e-setup-git e2e-setup-kind e2e-setup-fluxd
```

Makefile
```diff
@@ -173,3 +173,49 @@ server-profile-cpu:
...
+	- make e2e-do-release"
+
+e2e-teardown:
+	kind delete cluster
+
+e2e-setup-git:
+	rm -rf e2e-test/sourcegitrepo
+	mkdir -p e2e-test/sourcegitrepo/local/releases/default
+	echo "Hello World" > e2e-test/sourcegitrepo/README.md
+	cd e2e-test/sourcegitrepo;\
```
Crevil: The directory name is pretty hard to head like this. Could we maybe stick with kebab casing as with `e2e-test`?

```suggestion
	rm -rf e2e-test/source-git-repo
	mkdir -p e2e-test/source-git-repo/local/releases/default
	echo "Hello World" > e2e-test/source-git-repo/README.md
	cd e2e-test/source-git-repo;\
```

This has an effect on multiple targets so maybe put in into a variable as well?

Makefile
```diff
@@ -173,3 +173,49 @@ server-profile-cpu:
...
+	kind create cluster --config e2e-test/kind-cluster.yaml
+
+e2e-setup-fluxd:
+	kubectl apply -f e2e-test/fluxd.yaml
+
+e2e-run-local-daemon:
+	go run ./cmd/daemon start --environment local --release-manager-url http://localhost:10080
+
+e2e-run-local-manager:
+	go run ./cmd/server start --ssh-private-key ~/.ssh/id_rsa --config-repo file:///Users/Emil/Workspace/lunarway/release-manager/e2e-test/sourcegitrepo --http-port 10080
```
Crevil: I guess this shouldn't point to an absolute local Git repo?

e2e-test/fluxd.yaml
```diff
@@ -0,0 +1,166 @@
...
...RmE5SjdpaDdIRCs4VgpqT3IydmxaMnNuZmlUcmdJMnNCODF4NHlMcVRJcENkSXgwMmdnWEJxbTdaVTBTMWk1bUpVTHlzNlN0Z0R4SVFac0xyc2xzCnlCUDVvWnJFVjlsQnpFYSttUXF3NWlodG1JUnlIWDNadFZEYkQ5VUJKZ01MbVAvY3Q1YlJ4RjhXTFU5R3hWd2xob1Q0aFEKQ2pBRlJBUlpBeHczTmlUUVVReDhYUDgydnhjZmdwTDJrc1k2dFFwSWdNREovNjJrYjNSaVJocjBGTGlKaGw2WkNCd2hJcApjK05UVzVhcXNTRWF0OURRRWRYcmNuK2hzTEtOZGVoUHUxeGxmS3IvYzRGdEFBQUFnUURtcGwrMU9GdXFwaTQrYnI0d3EyCmtkY1AvRk9jQlZuWXpwODE4eG1TQzNnMlc5YXlrb2FCem9jaTFNaXBoQVRWbTB1RDBzWHN2YXEzYW9ESWhBQ2FNL2t6bjkKcmNGb0JqeDFNUXYxVWUrWEJHWWN6S2V2OHFzR1ZYaWpwSnhIRGdNZkhTWmJHcEVzNmFhanpyNlJjeW1YR20wdVI3UGxFNApWQXU0L2JtdXljR2dBQUFJRUE3UWl3WHdaQzlnUWY1NVlpUDNsM1dDU3JBUFJ5Qjd2WXQxQW1BeWplRDM5K2lNcGNrVDViClRLWUo5ckcraWsvWHZNSVJKM1cyYW9LREMzd25STWhOMGdvcHpWTU9ZMVlnV2Vid0ZJdWwvcjRTZW5Jb0l3TG5Vcm1TYjEKK0YrUzVZRWRMUHBibGZRa1kxdDlzbUNqUzFIcTVVVkZ3VGw0KzdpK3Z3OS8zNVRjc0FBQUNCQU5namtXelBMNHg4cHB5dwptY2YwMkxBMVVZbE5idEpSbGk5aktTRE1wOGtxNkJWSjh4SGwwcWVHNWwxYWkvaTAvaWxWd3ZYb1JKYnBuSThPUFlJNW43Cm0vZWtSd3doamtWWnBoMDNEbExtckc4dkZVZUNtOFZuT2tNQk00YmVZSnd1ZGhBOWMxSzdUcHJkNkRaaUFzNGxMMXBDQTUKeWZkWkJ2UVhRdC9qYXdkdkFBQUFIa1Z0YVd4QVJXMXBiSE10VFdGalFtOXZheTFRY204dE1pNXNiMk5oYkFFQ0F3UT0KLS0tLS1FTkQgT1BFTlNTSCBQUklWQVRFIEtFWS0tLS0tCg==
```
Crevil: Is this needed and in that case, where does it come from?

e2e-test/fluxd.yaml
```diff
@@ -0,0 +1,166 @@
...
+          image: fluxcd/flux-recv:0.3.0
+          imagePullPolicy: IfNotPresent
+          args:
+            - "--config=/etc/fluxrecv/fluxrecv.yaml"
+          ports:
+            - containerPort: 8080
+              hostPort: 8080
+          volumeMounts:
+            - name: fluxrecv-config
+              mountPath: /etc/fluxrecv
```
Crevil: `recv` won't work with hooks when running here, will it? In that case we might as well remove it entirely?

cmd/daemon/kubernetes/kubernetes.go
```diff
@@ -27,7 +29,15 @@ var (
...
-	config, err := rest.InClusterConfig()
+	var config *rest.Config
+	var err error
+	kubeconfig := os.Getenv("KUBECONFIG")
+	if kubeconfig == "" {
+		config, err = rest.InClusterConfig()
+	} else {
+		config, err = clientcmd.BuildConfigFromFlags("", kubeconfig)
+	}
+
```
Crevil: This could be simplified a bit as `BuildConfigFromFlags` [falls back to using `InClusterConfig`](https://godoc.org/k8s.io/client-go/tools/clientcmd#BuildConfigFromFlags).
```suggestion
	config, err := clientcmd.BuildConfigFromFlags("", os.Getenv("KUBECONFIG"))
```

cmd/daemon/kubernetes/kubernetes.go
```diff
@@ -27,7 +29,15 @@ var (
 )
 
 func NewClient() (*Client, error) {
```
Crevil: What do you think about moving the environment lookup out to a flag with a default value in `cmd/daemon/command/start.go` instead? Just to avoid having environment lookups in the code.

e2e-test/fluxd.yaml
```diff
@@ -0,0 +1,166 @@
...
+      labels:
+        name: flux
+        type: deployment
+    spec:
+      serviceAccountName: flux
+      volumes:
+        - name: git-key
+          secret:
+            secretName: flux-git-deploy
+            defaultMode: 0400 # when mounted read-only, we won't be able to chmodª
```
Crevil: This is not mounted on any container?

e2e-test/fluxd.yaml
```diff
@@ -0,0 +1,166 @@
...
+            - name: KUBECONFIG
+              value: /root/.kubectl/config
+          volumeMounts:
+            # - name: git-key
+            #   mountPath: /etc/fluxd/ssh
+            #   readOnly: true
+            - name: git-keygen
+              mountPath: /var/fluxd/keygen
+            - name: tmp
+              mountPath: /tmp
```
Crevil: This is not used if I read the `args` right?

Makefile
emilingerslev: does this run before "body" of the target and sequentially?

Makefile
emilingerslev: Yeah, that we could do. Will change that 👍

Makefile
emilingerslev: no it shouldn't 😅 will fix

emilingerslev: no it shouldn't 😅 will fix

cmd/daemon/kubernetes/kubernetes.go
emilingerslev: I love it. I just wanted to keep changes to an absolute minimum. Can add that change though. 👍

cmd/daemon/kubernetes/kubernetes.go
emilingerslev: Nice, didn't know that 💪

Makefile
Crevil: Yes.

e2e-test/fluxd.yaml
emilingerslev: It's needed in the current setup, just because it isn't optional. It is just generated on my machine. Will write a comment on it.

e2e-test/fluxd.yaml
emilingerslev: 🤔

e2e-test/fluxd.yaml
emilingerslev: Not sure. I stole it from our lunar-way-flux setup

e2e-test/fluxd.yaml
emilingerslev: Yeah, we could consider removing it. I was considering adding a fake webhook solution a some point

Crevil: With a sync interval of 5s it might not be that relevant. :) except if the events on the websocket are different on this trigger than of normal changes.

e2e-test/fluxd.yaml
Crevil: Ah yes. Flux does its operations on temporary directories, so this is still relevant 👍

e2e-test/fluxd.yaml
emilingerslev: Good enough?

e2e-test/fluxd.yaml
emilingerslev: Yeah, last part was also my thought. But lets try to remove it.
What about the release-manager, does it support pulling too? Or is it only webhookable?

e2e-test/fluxd.yaml
emilingerslev: It's a magic secret convention: https://github.com/fluxcd/flux/blob/a3319b09e362740963c7e971aa12ddb36387a24b/docs/guides/provide-own-ssh-key.md

I'm making it less hidden by using `--k8s-secret-name` flag

e2e-test/fluxd.yaml
```diff
@@ -0,0 +1,157 @@
...
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: flux-recv
+data:
+  fluxrecv.yaml: |
+    fluxRecvVersion: 1
+    endpoints: []
+---
```
Crevil: This CM could be deleted now, as `flux-recv` is no longer in the deployment?
```suggestion
```

e2e-test/fluxd.yaml
Crevil: Nice find 👍

e2e-test/fluxd.yaml
```diff
@@ -0,0 +1,157 @@
...
+      serviceAccountName: flux
+      volumes:
+        - name: git-key
+          secret:
+            secretName: flux-git-deploy
+            defaultMode: 0400 # when mounted read-only, we won't be able to chmodª
+        - name: fluxrecv-config
+          configMap:
+            name: flux-recv
+            defaultMode: 0400
```
Crevil: This can be removed as well now `flux-recv` is out?

```suggestion
```

e2e-test/fluxd.yaml
```diff
@@ -0,0 +1,157 @@
...
+              memory: 64Mi
+          ports:
+            - containerPort: 3030 # informational
+          env:
+            - name: KUBECONFIG
+              value: /root/.kubectl/config
+          volumeMounts:
+            # - name: git-key
+            #   mountPath: /etc/fluxd/ssh
+            #   readOnly: true
```
Crevil: Can this be removed also?

emilingerslev: definitely! 👍

e2e-test/fluxd.yaml
emilingerslev: definitely! 👍

e2e-test/fluxd.yaml
emilingerslev: definitely! 👍

PR #142, opened 2020-03-18
Add in-memory broker

Crevil:
This change set adds an in-memory broker implementation as a replacement for the
AMQP broker. This makes is simpler to run the server locally or in environments
where persistence of the queue is not necessary.

Package `amqp` is moved into package `broker` along with the `memory` implementation and the new `broker` package exports the interfaces required by a broker.

A new flag `broker-type` is added to configure what implementation is to be used
and the Make target server is replaced by `server-memory` and `server-rabbitmq`
configured respectively with the in-memory and AMQP brokers.

A small change to the contract between the publisher and consumer is added as
well. Event implementations are now required to implement both an `Unmarshal`
(former `Body`) method along with a `Marshal` method. This makes them self contained
and removes the odd `encoding/json` package dependency of package `amqp`.

The handler registration is also moved to the `StartConsumer` method instead of
being required at construction time. This makes for a simpler implementation of
brokers and also makes more sense in the domain of handlers being coupled with
consumers and not brokers in general.

## Future TODOs

I think a simple implementation of `Handlers` that is a `map[string]Handler` would make great sense as to decorate handlers with logging, panic handling etc. like `http.Handler`.
I've got some of the code ready but wouldn't want to clutter this PR up with the change right now.
When this is in place we can remove the duplicated log handling in the two broker implementations and replace them with a Handler middleware.

I'd also like to have a general broker test. `TestBroker_PublishAndConsumer` and `TestWorker_PublishAndConsumer` tests are basically the same and are in fact testing the the brokers are implementing the expected behaviour of a broker, eg. `StartConsumer` returns `broker.ErrBrokerClosed`.
I will properly add this in a later PR.


internal/flow/promote.go
```diff
@@ -115,8 +116,12 @@ func (PromoteEvent) Type() string {
...
 
-func (p PromoteEvent) Body() interface{} {
-	return p
+func (p PromoteEvent) Marshal() ([]byte, error) {
+	return json.Marshal(p)
+}
+
+func (p *PromoteEvent) Unmarshal(data []byte) error {
+	return json.Unmarshal(data, p)
 }
```
emilingerslev: Won't this get duplicated quite a lot? 🤔

Crevil: Yes. I think some sort of serializer layer would be nice. Like go-swagger har different producers maybe. 

But for the four events for now, it seemed ok.

It’s also what we do in our internal event package actually, now that I think of it.

emilingerslev: Alright, lets just continue, but i'd like to keep it on the consideration list 👍

Crevil: I've added it to the todo list. 👍

PR #148, opened 2020-03-30
Add authorEmail to commit message

kaspernissen:
This PR adds the e-mail of the git author to the commit message. 
This makes it easier to parse and determine the recipient when reporting flux events in Slack later on.


kaspernissen:
> Nice. Great to see all the added tests. 👏 

It’s primarily just a copy-paste with the addition for the author email. But I think it test the backwards compatibility we want

Crevil:
Yes, I figured, but that is also exactly what we want to ensure.

kaspernissen:
Yes, exactly. I'm cutting a release right now to get the new messages rolling.

PR #154, opened 2020-04-02
Refactor http in daemon

kaspernissen:
This PR switches the http client implementation in the daemon to our internal http client. 

Further, release-manager will not be returning unknown errors back to daemon, as it has resultet in quite a few errors in the daemon.

### DIFF
--- a/cmd/daemon/command/start.go
+++ b/cmd/daemon/command/start.go
@@ -16,16 +16,11 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func StartDaemon(version string) *cobra.Command {
+func StartDaemon() *cobra.Command {
 	var environment, kubeConfigPath, fluxApiBinding string
 	var logConfiguration *log.Configuration
 
-	client := httpinternal.Client{
-		Metadata: httpinternal.Metadata{
-			CLIVersion:  version,
-			CallerEmail: "daemon",
-		},
-	}
+	client := httpinternal.Client{}
 	var command = &cobra.Command{
 		Use:   "start",
 		Short: "start the release-daemon",
--- a/cmd/daemon/command/start.go
+++ b/cmd/daemon/command/start.go
@@ -1,10 +1,7 @@
 package command
 
 import (
-	"bytes"
 	"context"
-	"encoding/json"
-	"io/ioutil"
 	"net/http"
 	"os"
 	"os/signal"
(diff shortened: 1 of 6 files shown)


cmd/daemon/main.go
```diff
@@ -14,7 +14,7 @@ var (
 )
 
 func main() {
-	c, err := command.DaemonCommand()
+	c, err := command.DaemonCommand(&version)
```
emilingerslev, line 17: Why is this a pointer to a string?

cmd/daemon/flux/exporter.go
```diff
@@ -28,18 +22,19 @@ type Exporter interface {
...
-	client := &http.Client{
-		Timeout: 20 * time.Second,
+	var resp httpinternal.FluxNotifyResponse
+	url, err := f.Client.URL("webhook/daemon/flux")
+	if err != nil {
+		return err
 	}
-	b := &bytes.Buffer{}
-	err := json.NewEncoder(b).Encode(httpinternal.FluxNotifyRequest{
+	f.Log.Infof("URL: %s", url)
```
Crevil, line 36: Is this required to get logged? To me it looks a bit debuggy?

Crevil:
Niice and clean. Great to see more lines deleted than added 🎉

cmd/daemon/command/start.go
```diff
@@ -132,34 +145,11 @@ func notifyReleaseManager(event *kubernetes.PodEvent, logs, releaseManagerUrl, a
...
-		log.Errorf("error posting PodNotifyRequest to %s: %+v", url, err)
-		return
-	}
-	if resp.StatusCode != 200 {
-		body, err := ioutil.ReadAll(resp.Body)
-		if err != nil {
-			log.Errorf("failed to read response body: %+v", err)
-		}
-		log.Errorf("release-manager returned %s status-code in notify webhook: %s", resp.Status, body)
-		return
```
emilingerslev, line 161: nice with so much less 💪

cmd/daemon/flux/exporter.go
kaspernissen: It was a debug log line. It's removed 👍

kaspernissen:
I removed the versioning header, and CLI metadata in the client, as its very CLI minded. We could consider adding other relevant headers, but let's do that another time.

PR #155, opened 2020-04-02
Refactor flux event parsing into server

kaspernissen:
This PR moves the flux event handling to the release-manager server instead of the daemons

### DIFF
--- a/internal/flux/flux.go
+++ b/internal/flux/flux.go
@@ -1,0 +1,52 @@
+package flux
+
+import (
+	"github.com/weaveworks/flux/event"
+	"github.com/weaveworks/flux/update"
+)
+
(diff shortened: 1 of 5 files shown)


internal/flux/flux.go
```diff
@@ -0,0 +1,52 @@
...
+	case *event.AutoReleaseEventMetadata:
+		return v.Result
+	case *event.ReleaseEventMetadata:
+		return v.Result
+	default:
+		return update.Result{}
+	}
+}
+
+func GetChangedImages(meta event.EventMetadata) []string {
```
kaspernissen, line 34: I'm keeping these helpers here in case we need them going forward

PR #156, opened 2020-04-02
Remove the checkmark for k8s success 

kaspernissen:
the green bar already indicates success along with the message


kaspernissen:
Then we should consider doing it on all messages. Right now k8s message is the only one, along with what's coming from `artifact`. Then we should do it on all messages instead?

Crevil:
Yeah. We added it to artifact's messages a long time ago as some developers did not see the results very clearly.

kaspernissen:
Makes sense. I will close this one, and align messages in another PR 👍

PR #163, opened 2020-04-14
Add publish confirm mode for AMQP deliveries

Crevil:
This change set implements publish confirmation in the AMQP broker. It will
ensure that the AMQP server confirms all publishes and that the Publish method
does not return before this confirmation is received.

The timeout between publishes can be configured with flag
`amqp-republish-timeouts` and defaults to 1 second. During republishes contexts
cancellation and shutdowns are respected and will result in `context.Cancelled`
and `broker.ErrBrokerClosed` errors respectively.

As the current broker implementation does not handle channel failures, the test
case for this is commented for now. It does however document the "bad" behaviour
of not stopping or recreating the channel in this case. A future PR will fix
this as the failure mode is not introduced with this change.


Crevil:
Yes, until the publisher is closed or the context times out.

emilingerslev:
Make sense, because the only real case, is a problem with rabbitmq and not the message, right?

Crevil:
Yes. 👍

PR #166, opened 2020-04-15
Refactor and upgrade release-daemon

kaspernissen:
This PR refactors and upgrades the release-daemon in multiple ways. 

* Upgrade of the client-go and kubernetes libraries to 1.18.0
* Switch the implementation from being only pod based, to now also including knowledge of deployments.
* Use internal http client for all communication

What does the release-daemon do? 
* Listens for events from Flux, and passes these events on to the release-manager
* Watches the kubernetes api for successful deployments, e.g. deployments where all pods are rolled out, and reports the status back to the release-manager. 
* Watches pods for failures, currently looking for two specific cases: `CrashLoopBackOff` and `CreateContainerConfigError`. 
* If a `CreateContainerConfigError` is detected we extract the error message and generates a `PodErrorEvent` that is sent back to release-manager.
* If a `CrashLoopBackOff` is detected we extract logs of the given containers, and report them back to the release-manager. Further an option to mute the reporting using Modulo is added. Default to mod 5. `--modulo-crash-report-notif`.

What's missing? 
* StatefulSet reporting
* DaemonSet reporting
(should be similar to watching a Deployment)

This PR further changes the reporting of the release-manager in Slack to the following: 
Screenshot 2020-04-15 at 19 12 47


cmd/daemon/kubernetes/deployments.go
```diff
@@ -0,0 +1,144 @@
...
+	if deploy.Annotations["lunarway.com/author"] == "" {
+		log.Errorf("author missing in deployment: namespace '%s' pod '%s'", deploy.Namespace, deploy.Name)
+		return false
+	}
+	return true
+}
+
+// Avoid reporting on pods that has been marked for termination
+func isDeploymentMarkedForTermination(deploy *appsv1.Deployment) bool {
+	if deploy.DeletionTimestamp != nil {
```
golangcibot, line 140: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

cmd/daemon/kubernetes/pods.go
```diff
@@ -0,0 +1,212 @@
...
+	if pod.Annotations["lunarway.com/artifact-id"] == "" {
+		log.Errorf("artifact-id missing in pod template: namespace '%s' pod '%s'", pod.Namespace, pod.Name)
+		return false
+	}
+	return true
+}
+
+// Avoid reporting on pods that has been marked for termination
+func isPodMarkedForTermination(pod *corev1.Pod) bool {
+	if pod.DeletionTimestamp != nil {
```
golangcibot, line 135: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

cmd/daemon/kubernetes/deployments.go
```diff
@@ -0,0 +1,144 @@
...
+			return httpinternal.DeploymentEvent{}, false, nil
+		}
+
+		newRs, err := deploymentutil.GetNewReplicaSet(deployment, c.AppsV1())
+		if err != nil {
+			return httpinternal.DeploymentEvent{}, false, nil
+		}
+		log.Infof("New Replicaset: %s, CreationTime: %s", newRs.Name, newRs.CreationTimestamp)
+		//We discard events if the difference between creation and now is greater than 10s
+		diff := time.Now().Sub(newRs.CreationTimestamp.Time)
```
golangcibot, line 100: S1012: should use `time.Since` instead of `time.Now().Sub` (from `gosimple`)

cmd/daemon/kubernetes/deployments.go
```diff
@@ -0,0 +1,144 @@
...
+			err = c.exporter.SendSuccessfulDeploymentEvent(ctx, event)
+			if err != nil {
+				log.Errorf("error SendSuccessfulDeploymentEvent")
+				continue
+			}
+		case <-ctx.Done():
+			watcher.Stop()
+		}
+	}
+	return nil
```
golangcibot, line 66: unreachable: unreachable code (from `govet`)

cmd/daemon/kubernetes/deployments.go
```diff
@@ -0,0 +1,137 @@
...
```
golangcibot: unreachable: unreachable code (from `govet`)

cmd/daemon/kubernetes/kubernetes.go
```diff
@@ -1,332 +1,58 @@
...
-	podLogOpts := v1.PodLogOptions{
-		TailLines: &numberOfLogLines,
-	}
-	req := c.clientset.CoreV1().Pods(namespace).GetLogs(podName, &podLogOpts)
-
-	podLogs, err := req.Stream()
-	if err != nil {
-		return "", err
+func NewClient(kubeConfigPath string, moduloCrashReportNotif float64, replicaSetTimeDiff time.Duration, e Exporter) (*Client, error) {
+	if kubeConfigPath != "" {
```
Crevil, line 26: There is no need to handle this case manually. `BuildConfigFromFlags` will fallback to `InClusterConfig` if the path is empty.
https://godoc.org/k8s.io/client-go/tools/clientcmd#BuildConfigFromFlags

cmd/daemon/kubernetes/pods.go
```diff
@@ -0,0 +1,209 @@
...
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"math"
+	"strings"
+
+	httpinternal "github.com/lunarway/release-manager/internal/http"
```
Crevil, line 12: Again, no need for the import alias here

```suggestion
	"github.com/lunarway/release-manager/internal/http"
```

cmd/daemon/kubernetes/pods.go
```diff
@@ -0,0 +1,209 @@
...
+				})
+				if err != nil {
+					log.Errorf("error SendCreateContainerConfigError")
+				}
+				continue
+			}
+
+		case <-ctx.Done():
+			watcher.Stop()
+			return ctx.Err()
```
Crevil, line 114: Regarding the placement, see the deployment handler's comment.

In regards to the returned error, I think the way we do it in the deployment handler is better than this.

This will give a `context.Cancel` error when stopping cancelling the context instead of `ErrWatcherClosed`. I would expect `ErrWatcherClosed` and in the calling logic check `ctx.Err()` first as to know if the watcher closed unexpectedly or the caller cancelled it.
```suggestion
```

cmd/daemon/kubernetes/pods_test.go
```diff
@@ -0,0 +1,50 @@
+package kubernetes
+
+import (
+	"testing"
+
+	"github.com/pkg/errors"
+
```
Crevil, line 7:
```suggestion
```

cmd/server/http/http.go
```diff
@@ -50,8 +50,9 @@ func NewServer(opts *Options, slackClient *slack.Client, flowSvc *flow.Service,
 	mux.HandleFunc("/policies/", trace(tracer, authenticate(opts.HamCtlAuthToken, policy(&payloader, policySvc))))
 	mux.HandleFunc("/describe/", trace(tracer, authenticate(opts.HamCtlAuthToken, describe(&payloader, flowSvc))))
 	mux.HandleFunc("/webhook/github", trace(tracer, githubWebhook(&payloader, flowSvc, policySvc, gitSvc, slackClient, opts.GithubWebhookSecret)))
-	mux.HandleFunc("/webhook/daemon", trace(tracer, authenticate(opts.DaemonAuthToken, daemonWebhook(&payloader, flowSvc))))
```
Crevil, line 53: This is a breaking change in case you are unaware. Daemons will fail during rollout until both the release manager and daemons are up on the new version.

I don't think its a problem here as we control both apps, but we cannot keep the old daemon running in some of the environments.

We could keep the old endpoint along with the handler maybe? To give us selves some more air during the deployment for testing it in only dev?

cmd/server/http/http.go
```diff
@@ -342,66 +343,84 @@ func rollback(payload *payload, flowSvc *flow.Service) http.HandlerFunc {
...
-		err = flowSvc.NotifyFluxEvent(ctx, &fluxNotifyEvent)
+func daemonk8sPodErrorWebhook(payload *payload, flowSvc *flow.Service) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		// copy span from request context but ignore any deadlines on the request context
+		ctx := opentracing.ContextWithSpan(context.Background(), opentracing.SpanFromContext(r.Context()))
+		logger := log.WithContext(ctx)
+		var event httpinternal.PodErrorEvent
+		err := payload.decodeResponse(ctx, r.Body, &event)
+		if err != nil {
+			logger.Errorf("http: daemon k8s create config error webhook: decode request body failed: %v", err)
```
Crevil, line 409:
```suggestion
			logger.Errorf("http: daemon k8s pod error webhook: decode request body failed: %v", err)
```

cmd/server/http/http.go
```diff
@@ -342,66 +343,84 @@ func rollback(payload *payload, flowSvc *flow.Service) http.HandlerFunc {
...
+			invalidBodyError(w)
+			return
+		}
+		logger = logger.WithFields("event", event)
+		err = flowSvc.NotifyK8SPodErrorEvent(ctx, &event)
 		if err != nil && errors.Cause(err) != slack.ErrUnknownEmail {
-			logger.Errorf("http: daemon flux webhook failed: %+v", err)
+			logger.Errorf("http: daemon k8s deploy webhook failed: %+v", err)
```
Crevil, line 416:
```suggestion
			logger.Errorf("http: daemon k8s pod error webhook failed: %+v", err)
```

cmd/server/http/http.go
```diff
@@ -342,66 +343,84 @@ func rollback(payload *payload, flowSvc *flow.Service) http.HandlerFunc {
...
 		}
 		w.WriteHeader(http.StatusOK)
-		err = payload.encodeResponse(ctx, w, httpinternal.PodNotifyResponse{})
+		err = payload.encodeResponse(ctx, w, httpinternal.KubernetesNotifyResponse{})
 		if err != nil {
-			logger.Errorf("http: daemon flux webhook: environment: '%s' marshal response: %v", fluxNotifyEvent.Environment, err)
+			logger.Errorf("http: daemon k8s deploy webhook: environment: '%s' marshal response: %v", event.Environment, err)
```
Crevil, line 421:
```suggestion
			logger.Errorf("http: daemon k8s pod error webhook: environment: '%s' marshal response: %v", event.Environment, err)
```

cmd/server/http/http.go
```diff
@@ -342,66 +343,84 @@ func rollback(payload *payload, flowSvc *flow.Service) http.HandlerFunc {
...
 		}
-		logger.Infof("http: daemon flux webhook: handled")
+		logger.Infof("http: daemon k8s deploy webhook: handled")
```
Crevil, line 423:
```suggestion
		logger.Infof("http: daemon k8s pod error webhook: handled")
```

go.mod
```diff
@@ -49,9 +44,57 @@ require (
...
+	k8s.io/legacy-cloud-providers v0.0.0
+	k8s.io/metrics v0.18.0
+	k8s.io/sample-apiserver v0.0.0
+)
+
+// k8s.io/kubernetes has a go.mod file that sets the version of the following
+// modules to v0.0.0. This causes go to throw an error. These need to be set
+// to a version for Go to process them. Here they are set to the same
+// revision as the marked version of Kubernetes. When Kubernetes is updated
+// these need to be updated as well.
```
Crevil, line 76: Do we have a source of this method? Just for future reference in understanding it?

internal/http/types.go
```diff
@@ -87,9 +87,37 @@ type FluxNotifyRequest struct {
 	FluxEvent   event.Event
 }
 
+type DeploymentEvent struct {
+	Name          string `json:"name,omitempty"`
+	Namespace     string `json:"namespace,omitempty"`
+	AvailablePods int32  `json:"podCount,omitempty"`
```
Crevil, line 93: For consistency could we use `availablePods` as the JSON field? Further I think we should not omit empty values (`0`) just to better understand the payloads.

```suggestion
	AvailablePods int32  `json:"availablePods"`
```

internal/http/types.go
```diff
@@ -87,9 +87,37 @@ type FluxNotifyRequest struct {
...
+	Replicas      int32  `json:"replicas,omitempty"`
+	ArtifactID    string `json:"artifactId,omitempty"`
+	AuthorEmail   string `json:"authorEmail,omitempty"`
+	Environment   string `json:"environment,omitempty"`
+}
+type ContainerError struct {
+	Name         string `json:"name"`
+	ErrorMessage string `json:"message"`
+	Type         string `json:"type"`
```
Crevil, line 102: We could omit empty values here.

```suggestion
	Name         string `json:"name,omitempty"`
	ErrorMessage string `json:"message,omitempty"`
	Type         string `json:"type,omitempty"`
```

internal/http/types.go
```diff
@@ -87,9 +87,37 @@ type FluxNotifyRequest struct {
...
+}
+
+type PodErrorEvent struct {
+	PodName     string `json:"podName,omitempty"`
+	Namespace   string `json:"namespace,omitempty"`
+	Errors      []ContainerError
```
Crevil, line 108: Missing a JSON tag

```suggestion
	Errors      []ContainerError `json:"errors,omitempty"`
```

internal/http/types.go
```diff
@@ -87,9 +87,37 @@ type FluxNotifyRequest struct {
...
+	AuthorEmail string `json:"authorEmail,omitempty"`
+	Environment string `json:"environment,omitempty"`
+	ArtifactID  string `json:"artifactId,omitempty"`
+}
+type KubernetesNotifyResponse struct {
+}
+
+//DEPRECATED
```
Crevil, line 116: Are these deprecated types still used?

kaspernissen:
> That sure is a major upgrade. Overall this looks super good. There is a lot going on but you got some great function names thrown around to ease with understanding all the cases. Great job 🎉
> 
> As usual I have a lot of comments. Most of them are minor details but let me know if something is unclear.

As expected 😉 I will go through them and request review afterwards 👍

cmd/daemon/kubernetes/deployments.go
```diff
@@ -0,0 +1,136 @@
...
+				continue
+			}
+		case <-ctx.Done():
+			watcher.Stop()
+		}
+	}
+}
+
+func isDeploymentSuccessful(c *kubernetes.Clientset, replicaSetTimeDiff time.Duration, deployment *appsv1.Deployment) (httpinternal.DeploymentEvent, bool, error) {
+	if deployment.Generation <= deployment.Status.ObservedGeneration {
```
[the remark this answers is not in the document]
kaspernissen: I'm not sure I follow. You want the happy path to come first? We can probably simplify this using some deploymentutil functions, as these if statements are exactly what kubectl does.

cmd/daemon/kubernetes/deployments.go
```diff
@@ -0,0 +1,136 @@
...
+		cond := deploymentutil.GetDeploymentCondition(deployment.Status, appsv1.DeploymentProgressing)
+		if cond != nil && cond.Reason == "ProgressDeadlineExceeded" {
+			log.Errorf("deployment %q exceeded its progress deadline", deployment.Name)
```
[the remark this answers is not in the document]
kaspernissen: It's the same log lines, as was previous printed in kubectl. We can just use `%s`

cmd/daemon/kubernetes/deployments.go
```diff
@@ -0,0 +1,136 @@
...
+			// TODO: Maybe return a specific error here
```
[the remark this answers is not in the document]
kaspernissen: That's right. Usually, this happens when a pod is in `CreateContainerConfigError` which we will catch. 

We could, however, as the TODO suggest, maybe return some specific error that can report an error.

cmd/daemon/kubernetes/deployments.go
```diff
@@ -0,0 +1,136 @@
...
+			return httpinternal.DeploymentEvent{}, false, nil
+		}
+		// Waiting for deployment %q rollout to finish: %d old replicas are pending termination...
+		if deployment.Status.Replicas > deployment.Status.UpdatedReplicas {
+			return httpinternal.DeploymentEvent{}, false, nil
+		}
+		// Waiting for deployment %q rollout to finish: %d of %d updated replicas are available...
+		if deployment.Status.AvailableReplicas < deployment.Status.UpdatedReplicas {
+			return httpinternal.DeploymentEvent{}, false, nil
+		}
```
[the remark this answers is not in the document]
kaspernissen: This is an exact copy of how kubectl does it. As mentioned in a comment earlier, we could look into finding this functionality in deployment utils.

cmd/daemon/kubernetes/deployments.go
```diff
@@ -0,0 +1,136 @@
...
+
+		newRs, err := deploymentutil.GetNewReplicaSet(deployment, c.AppsV1())
+		if err != nil {
+			return httpinternal.DeploymentEvent{}, false, nil
+		}
+		//We discard events if the difference between creation time of the ReplicaSet and now is greater than replicaSetTimeDiff
+		diff := time.Since(newRs.CreationTimestamp.Time)
+		if diff > (replicaSetTimeDiff) {
```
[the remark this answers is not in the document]
kaspernissen: leftovers

cmd/daemon/kubernetes/kubernetes.go
kaspernissen: Ah, I didn't know. That's awesome 🎉 So much easier!

cmd/server/http/http.go
kaspernissen: Do you think it's worth the work to support this? @Crevil

cmd/daemon/kubernetes/deployments.go
```diff
@@ -0,0 +1,128 @@
...
+	return http.DeploymentEvent{
+		Name:          deployment.Name,
+		Namespace:     deployment.Namespace,
+		ArtifactID:    deployment.Annotations["lunarway.com/artifact-id"],
+		AuthorEmail:   deployment.Annotations["lunarway.com/author"],
+		AvailablePods: deployment.Status.AvailableReplicas,
+		Replicas:      *deployment.Spec.Replicas,
+	}, true, nil
+
+	return http.DeploymentEvent{}, false, nil
```
golangcibot, line 103: unreachable: unreachable code (from `govet`)

kaspernissen:
Allright, I think I got through the comments now. Thanks @Crevil!

I switched out the logic in `isDeploymentSuccessful` with the function `DeploymentComplete` from the deployment utils package, with some of the custom logic after this check

kaspernissen:
The biggest question is probably whether or not we want to keep some backwards compatability or not? Is it worth it as we are controlling both applications?

Crevil:
Your call. It will make it a complete rollout to all environments, but it is fine with me. :)

kaspernissen:
Let's get this merged - and then fix reporting for `DaemonSet` and `StatefulSet` in another PR, and then just roll everything out in one go.

cmd/daemon/kubernetes/pods.go
```diff
@@ -0,0 +1,207 @@
...
+
+				err = c.exporter.SendPodErrorEvent(ctx, http.PodErrorEvent{
+					PodName:     pod.Name,
+					Namespace:   pod.Namespace,
+					Errors:      errorContainers,
+					ArtifactID:  pod.Annotations["lunarway.com/artifact-id"],
+					AuthorEmail: pod.Annotations["lunarway.com/author"],
+				})
+				if err != nil {
+					log.Errorf("error SendCrashLoopBackOffEvent")
```
Crevil, line 82: The error is missing from this log.
```suggestion
					log.Errorf("Failed to send crash loop backoff event: %v", err)
```

cmd/daemon/kubernetes/pods.go
```diff
@@ -0,0 +1,207 @@
...
+				if math.Mod(float64(restartCount), c.moduloCrashReportNotif) != 1 {
+					continue
+				}
+				var errorContainers []http.ContainerError
+				for _, cst := range pod.Status.ContainerStatuses {
+					if cst.State.Waiting != nil && cst.State.Waiting.Reason == "CrashLoopBackOff" {
+
+						logs, err := getLogs(ctx, c.clientset, pod.Name, cst.Name, pod.Namespace)
+						if err != nil {
+							log.Errorf("Error retrieving logs from pod: %s, container: %s, namespace: %s", pod.Name, cst.Name, pod.Namespace)
```
Crevil, line 63: The error is missing from this log

```suggestion
							log.Errorf("Error retrieving logs from pod: %s, container: %s, namespace: %s: %v", pod.Name, cst.Name, pod.Namespace, err)
```

cmd/daemon/kubernetes/pods.go
```diff
@@ -0,0 +1,207 @@
...
+
+				err = c.exporter.SendPodErrorEvent(ctx, http.PodErrorEvent{
+					PodName:     pod.Name,
+					Namespace:   pod.Namespace,
+					Errors:      errorContainers,
+					ArtifactID:  pod.Annotations["lunarway.com/artifact-id"],
+					AuthorEmail: pod.Annotations["lunarway.com/author"],
+				})
+				if err != nil {
+					log.Errorf("error SendCreateContainerConfigError")
```
Crevil, line 109: The error is missing from this log

```suggestion
					log.Errorf("Failed to send create container config error: %v", err)
```

cmd/daemon/kubernetes/deployments.go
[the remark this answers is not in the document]
Crevil: Current implementation is like this

```go
func isDeploymentSuccessful(c *kubernetes.Clientset, replicaSetTimeDiff time.Duration, deployment *appsv1.Deployment) (httpinternal.DeploymentEvent, bool, error) {
	if deployment.Generation <= deployment.Status.ObservedGeneration {
		// do a lot of stuff in 70 lines
		return httpinternal.DeploymentEvent{/* values */}, true, nil
	}
	return httpinternal.DeploymentEvent{}, false, nil
}
```

This moves the uninteresting case where observed generation is over deployment generation far apart.

Usually we use early returns and guards, like you do all over the place actually :) 


```go
func isDeploymentSuccessful(c *kubernetes.Clientset, replicaSetTimeDiff time.Duration, deployment *appsv1.Deployment) (httpinternal.DeploymentEvent, bool, error) {
	if deployment.Generation > deployment.Status.ObservedGeneration {
		return httpinternal.DeploymentEvent{}, false, nil
	}
	// do a lot of stuff in 70 lines
	return httpinternal.DeploymentEvent{/* values */}, true, nil
}
```

PR #167, opened 2020-04-16
Add global branch restriction policies

Crevil:
This PR adds the concept of global policies on the release-manager `server`. It
adds a `policy-branch-restrictions` flag to the server that can be used to limit
releases to specific environments across all services.

The policies are not different from those that can be set by `hamctl policy apply`
and can be inspected with `hamctl policy list` as usual.

When applying new policies we check that they do not
conflict with existing ones, ie. an `auto-release` policy with a non-compatible
branch restricted by a `branch-restriction` policy.

# Implementation

A `GlobalBranchRestrictionPolicies` field is added to `policy.Service` and policies are
merged with service specific policies in the `Get` method. This makes them appear
just like service specific policies but without an `ID` to indicate that they
cannot be deleted with `hamctl policy delete`.

Method `ApplyBranchRestriction`is updated to validate new `branch-restriction`
policies against the globally configured policies.

In order to write a test of `ApplyBranchRestriction` I changed the `git.Service`
dependency to an interface and with that changed the `MasterPath` string field of
`git.Service` to a metod. This makes it possible to mock the implementation and
test it without having git specific things in the test.

# Others

I changed the `NewStart` function to have a struct with all the options instead of
the ever increasing list of arguments.


internal/policy/policy.go
```diff
@@ -95,6 +110,36 @@ func (s *Service) Get(ctx context.Context, svc string) (Policies, error) {
 	return policies, nil
 }
 
+func mergeBranchRestrictions(ctx context.Context, svc string, global, local []BranchRestriction) []BranchRestriction {
+	if len(global) == 0 {
+		return local
+	}
+	var branchRestrictions []BranchRestriction
+	for _, globalRestriction := range global {
```
golangcibot, line 118: S1011: should replace loop with `branchRestrictions = append(branchRestrictions, global...)` (from `gosimple`)

PR #168, opened 2020-04-16
Watch DaemonSets and StatefulSets for new revisions

kaspernissen:
This PR adds functionality to the release-daemon to watch for successful releases. 

Only new releases will trigger a notification. Pod deletion or restarts will not trigger a notification as it would have previously. 

In order to control this behavior, the release-daemon adds an annotation to the `DaemonSets` and `StatefulSets` it's watching called `observed-artifact-id`. This is used to determine when a notification should trigger.

This PR also changes the `DeploymentEvent` to a more generic `ReleaseEvent` that can we used for all the different release types. The most significant change is the addition of a `Type`  that will be outputtet in the Slack to indicate which resource type the message is for. We can discuss whether or not this is relevant for developers.

This PR also changes the deployment watch to use the `observed-artifact-id` annotation.

### DIFF
--- a/cmd/daemon/kubernetes/daemonsets.go
+++ b/cmd/daemon/kubernetes/daemonsets.go
@@ -100,15 +100,9 @@ func isDaemonSetSuccessful(ds *appsv1.DaemonSet) bool {
 }
 
 func isDaemonSetCorrectlyAnnotated(ds *appsv1.DaemonSet) bool {
-	if !(ds.Annotations["lunarway.com/controlled-by-release-manager"] == "true") {
-		return false
-	}
-	if ds.Annotations["lunarway.com/artifact-id"] == "" {
-		log.Errorf("artifact-id missing in deployment: namespace '%s' name '%s'", ds.Namespace, ds.Name)
-		return false
-	}
-	if ds.Annotations["lunarway.com/author"] == "" {
-		log.Errorf("author missing in deployment: namespace '%s' name '%s'", ds.Namespace, ds.Name)
+	if !(ds.Annotations["lunarway.com/controlled-by-release-manager"] == "true") &&
+		ds.Annotations["lunarway.com/artifact-id"] == "" &&
+		ds.Annotations["lunarway.com/author"] == "" {
 		return false
 	}
 	return true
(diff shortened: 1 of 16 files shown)


cmd/daemon/kubernetes/daemonsets.go
```diff
@@ -0,0 +1,132 @@
...
+	return true
+}
+
+// Avoid reporting on pods that has been marked for termination
+func isDaemonSetMarkedForTermination(ds *appsv1.DaemonSet) bool {
+	return ds.DeletionTimestamp != nil
+}
+
+func isDaemonSetMarkedAsReleased(ds *appsv1.DaemonSet) bool {
+	if ds.Annotations["lunarway.com/released"] == "" {
```
golangcibot, line 117: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

cmd/daemon/kubernetes/daemonsets.go
```diff
@@ -0,0 +1,132 @@
...
+			log.Infof("Observed: %d, Gen: %d", ds.Status.ObservedGeneration, ds.Generation)
+
+			// Verify if the DaemonSet fulfills the criterias for a succesful release
+			ok = isDaemonSetSuccessful(ds)
+			if !ok {
+				continue
+			}
+
+			// Check if the DaemonSet is marked with the lunarway.com/released annotation
+			if isDaemonSetMarkedAsReleased(ds) {
```
kaspernissen, line 57: @Crevil @emilingerslev Any ideas on how to use the annotation to detect the new deployments that are applied instead of replaced? 

This will skip updates on all new apply's as the annotation will stay. 

We could maybe use the annotation `lunarway.com/artifact-id` that we already manages in this project? If that value is changed, it's a new release?

kaspernissen:
Moved the handling of checking annotations to a generic function `isCorrectlyAnnotated`. 

Further, change the order of the logic to first annotate after a successful message to the server. 

We could probably make the annotate function generic as well, but as @Crevil points out, it's probably not worth it.

PR #171, opened 2020-04-23
Add commit signing in artifact

kaspernissen:
This PR adds functionality to sign commits in `artifact` when pushing to the config repo. 

We utilize native git, and a `$HOME/.gitconfig` is required with the following configuration:

```
[user]
      name = Foo Bar
      email = test@lunar.app
      signingkey = signingkey
[commit]
	gpgsign = true
```

Further, this PR changes how the author is extracted in the `http`endpoint for the github webhook. As we now longer impersonate the user, an additional message is added to commit message: `Artifact-created-by: $authorName <$authorEmail>` This is similar to the way `--signoff` works, but we can't leverage this functionality here, as we can't read the individual users gitconfig in our pipeline.

*What does it do?*
As mentioned above, the current implementation impersonates the original git author on the source repo. We can't do that anymore when we want commits to be signed. In this PR we ensure that all commits originating from `artifact` are signed. This enables us to require signed commits, and minimizes this attack vector. Also impersonating users, is not best practice. With this implementation there intent is registered as an additional message to the commit message.

### DIFF
--- a/cmd/server/http/http.go
+++ b/cmd/server/http/http.go
@@ -453,6 +453,7 @@ func githubWebhook(payload *payload, flowSvc *flow.Service, policySvc *policyint
 			if err != nil {
 				logger.Infof("http: github webhook: extract author details from commit failed: message '%s'", payload.HeadCommit.Message)
 				w.WriteHeader(http.StatusOK)
+				return
 			}
 
 			// locate branch of commit. Look at both modified and added commits to
--- a/cmd/server/http/http.go
+++ b/cmd/server/http/http.go
@@ -451,7 +451,7 @@ func githubWebhook(payload *payload, flowSvc *flow.Service, policySvc *policyint
 			}
 			commitInfo, err := extractAuthorFromCommit(payload.HeadCommit.Message)
 			if err != nil {
-				logger.Infof("http: github webhook: no regex match from commit '%s'", payload.HeadCommit.Message)
+				logger.Infof("http: github webhook: extract author details from commit failed: message '%s'", payload.HeadCommit.Message)
 				w.WriteHeader(http.StatusOK)
 			}
 
--- a/cmd/server/http/http.go
+++ b/cmd/server/http/http.go
@@ -449,70 +449,68 @@ func githubWebhook(payload *payload, flowSvc *flow.Service, policySvc *policyint
 				w.WriteHeader(http.StatusOK)
 				return
 			}
-			rgx := regexp.MustCompile(`\[(.*?)\]`)
-			matches := rgx.FindStringSubmatch(payload.HeadCommit.Message)
-			if len(matches) < 2 {
-				logger.Infof("http: github webhook: no service match from commit '%s'", payload.HeadCommit.Message)
+			commitInfo, err := extractAuthorFromCommit(payload.HeadCommit.Message)
+			if err != nil {
+				logger.Infof("http: github webhook: no regex match from commit '%s'", payload.HeadCommit.Message)
 				w.WriteHeader(http.StatusOK)
-				return
 			}
-			serviceName := matches[1]
 
 			// locate branch of commit. Look at both modified and added commits to
 			// cover both updated artifacts and added ones (new versions vs first
 			// version)
-			branch, ok := git.BranchName(append(payload.HeadCommit.Added, payload.HeadCommit.Modified...), flowSvc.ArtifactFileName, serviceName)
+			branch, ok := git.BranchName(append(payload.HeadCommit.Added, payload.HeadCommit.Modified...), flowSvc.ArtifactFileName, commitInfo.Service)
 			if !ok {
-				logger.Infof("http: github webhook: service '%s': branch name not found", serviceName)
+				logger.Infof("http: github webhook: service '%s': branch name not found", commitInfo.Service)
 				w.WriteHeader(http.StatusOK)
 				return
 			}
 
-			logger = logger.WithFields("branch", branch, "service", serviceName, "commit", payload.HeadCommit)
+			logger = logger.WithFields("branch", branch, "service", commitInfo.Service, "commit", payload.HeadCommit)
 			// lookup policies for branch
-			autoReleases, err := policySvc.GetAutoReleases(ctx, serviceName, branch)
+			autoReleases, err := policySvc.GetAutoReleases(ctx, commitInfo.Service, branch)
 			if err != nil {
-				logger.Errorf("http: github webhook: service '%s' branch '%s': get auto release policies failed: %v", serviceName, branch, err)
-				err := slackClient.NotifySlackPolicyFailed(ctx, payload.HeadCommit.Author.Email, ":rocket: Release Manager :no_entry:", fmt.Sprintf("Auto release policy failed for service %s and %s", serviceName, branch))
+				logger.Errorf("http: github webhook: service '%s' branch '%s': get auto release policies failed: %v", commitInfo.Service, branch, err)
+				err := slackClient.NotifySlackPolicyFailed(ctx, commitInfo.AuthorEmail, ":rocket: Release Manager :no_entry:", fmt.Sprintf("Auto release policy failed for service %s and %s", commitInfo.Service, branch))
 				if err != nil {
 					logger.Errorf("http: github webhook: get auto-release policies: error notifying slack: %v", err)
 				}
 				unknownError(w)
 				return
 			}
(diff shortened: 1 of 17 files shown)


cmd/server/http/http.go
```diff
@@ -449,70 +449,68 @@ func githubWebhook(payload *payload, flowSvc *flow.Service, policySvc *policyint
...
 			}
-			rgx := regexp.MustCompile(`\[(.*?)\]`)
-			matches := rgx.FindStringSubmatch(payload.HeadCommit.Message)
-			if len(matches) < 2 {
-				logger.Infof("http: github webhook: no service match from commit '%s'", payload.HeadCommit.Message)
+			commitInfo, err := extractAuthorFromCommit(payload.HeadCommit.Message)
+			if err != nil {
+				logger.Infof("http: github webhook: no regex match from commit '%s'", payload.HeadCommit.Message)
 				w.WriteHeader(http.StatusOK)
-				return
```
Crevil, line 457: This return statement should stay, right? If we cannot parse the commit message to a meaningful service and author we do not know what to do?

cmd/server/http/http.go
```diff
@@ -449,70 +449,68 @@ func githubWebhook(payload *payload, flowSvc *flow.Service, policySvc *policyint
 				w.WriteHeader(http.StatusOK)
 				return
 			}
-			rgx := regexp.MustCompile(`\[(.*?)\]`)
-			matches := rgx.FindStringSubmatch(payload.HeadCommit.Message)
-			if len(matches) < 2 {
-				logger.Infof("http: github webhook: no service match from commit '%s'", payload.HeadCommit.Message)
+			commitInfo, err := extractAuthorFromCommit(payload.HeadCommit.Message)
+			if err != nil {
+				logger.Infof("http: github webhook: no regex match from commit '%s'", payload.HeadCommit.Message)
```
Crevil, line 454:
```suggestion
				logger.Infof("http: github webhook: extract author details from commit failed: message '%s'", payload.HeadCommit.Message)
```

cmd/server/http/http.go
```diff
@@ -728,3 +726,26 @@ func release(payload *payload, flowSvc *flow.Service) http.HandlerFunc {
...
 }
+
+type commitInfo struct {
+	AuthorName  string
+	AuthorEmail string
+	Service     string
+}
+
+func extractAuthorFromCommit(message string) (commitInfo, error) {
+	pattern := `^\[(?P<service>.*)\].*\nSigned-off-by:\s(?P<authorName>.*)\s<(?P<authorEmail>.*)>`
```
Crevil, line 737: This could be compiled ahead of time to reduce the need for it on every webhook.

cmd/server/http/http.go
```diff
@@ -728,3 +726,26 @@ func release(payload *payload, flowSvc *flow.Service) http.HandlerFunc {
...
+	r, err := regexp.Compile(pattern)
+	if err != nil {
+		return commitInfo{}, errors.WithMessage(err, "regex didn't match")
```
Crevil, line 740: This does in fact not decide if it matches but only verify the regular expression is valid and compilable.

If we compile it ahead of time this check can be removed.

cmd/server/http/http.go
```diff
@@ -728,3 +726,26 @@ func release(payload *payload, flowSvc *flow.Service) http.HandlerFunc {
...
+	}
+	matches := r.FindStringSubmatch(message)
+	if len(matches) < 1 {
+		return commitInfo{}, errors.New("not enough matches")
+	}
```
Crevil, line 745: [`FindStringSubmatch`](https://godoc.org/regexp#Regexp.FindStringSubmatch) returns `nil` in case of no matches so maybe the check should just be that?
As I can see from your tests, it does not look like it will ever return a partial match, so commits missing the `Signed-off-by` line will also result in `nil`.

```suggestion
	matches := r.FindStringSubmatch(message)
	if matches == nil {
		return commitInfo{}, errors.New("no match")
	}
```

cmd/server/http/http.go
```diff
@@ -728,3 +726,26 @@ func release(payload *payload, flowSvc *flow.Service) http.HandlerFunc {
...
 	return t.UnixNano() / int64(time.Millisecond)
 }
+
+type commitInfo struct {
+	AuthorName  string
+	AuthorEmail string
+	Service     string
+}
+
+func extractAuthorFromCommit(message string) (commitInfo, error) {
```
Crevil, line 736: You use two names for this function. `extractAuthorFromCommit` and in the test its `commitMessageExtraction`.
I think the latter is more precise, but maybe something like `extractInfoFromCommit` returning `commitInfo` reads well?

cmd/server/http/http_test.go
```diff
@@ -72,3 +73,42 @@ func TestAuthenticate(t *testing.T) {
 		})
 	}
 }
+
+func TestCommitMessageExtraction(t *testing.T) {
```
Crevil, line 77: Should be aligned with the tested function's name. :)

cmd/server/http/http_test.go
```diff
@@ -72,3 +73,42 @@ func TestAuthenticate(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			info, err := extractAuthorFromCommit(tc.commitMessage)
+			if tc.err != nil {
+				assert.EqualError(t, errors.Cause(err), tc.err.Error(), "output error not as expected")
+			} else {
+				assert.NoError(t, err, "no output error expected")
+			}
+			assert.Equal(t, tc.commitInfo.AuthorName, info.AuthorName, "AuthorName not as expected")
+			assert.Equal(t, tc.commitInfo.Service, info.Service, "Service not as expected")
+			assert.Equal(t, tc.commitInfo.AuthorEmail, info.AuthorEmail, "AuthorEmail not as expected")
```
Crevil, line 111: If we in the future add additional fields to `commitInfo` they will not automatically be asserted for with this approach.
It could be reduced to just comparing the `commitInfo` structs directly and any new fields then gets included automatically.

```suggestion
			assert.Equal(t, tc.commitInfo, info, "commit info not as expected")
```

internal/git/git.go
```diff
@@ -393,6 +393,40 @@ func (s *Service) Commit(ctx context.Context, rootPath, changesPath, authorName,
 	return gitPush(ctx, rootPath)
 }
 
+func (s *Service) SignedCommit(ctx context.Context, rootPath, changesPath, authorName, authorEmail, msg string, committer Committer) error {
```
Crevil, line 396: Argument `committer` is no longer used, right?

internal/git/git.go
```diff
@@ -393,6 +393,40 @@ func (s *Service) Commit(ctx context.Context, rootPath, changesPath, authorName,
...
+		return errors.WithMessage(err, "add changes")
+	}
+
+	span, _ = s.Tracer.FromCtx(ctx, "check for changes")
+	err = checkStatus(ctx, rootPath)
+	span.Finish()
+	if err != nil {
+		return errors.WithMessage(err, "check for changes")
+	}
+	fullCommitMsg := fmt.Sprintf("%s\nSigned-off-by: %s <%s>", msg, authorName, authorEmail)
```
Crevil, line 413: Have you tried to do an artifact release from an artifact created with this commit message?

We should make sure the locate conditions will still detect the releases properly, e.g. `locateReleaseCondition` and `locateServiceReleaseCondition`.

internal/git/git.go
```diff
@@ -564,11 +598,17 @@ func userHomeDir() string {
 	}
 }
 
+type Committer struct {
+	Name       string
+	Email      string
+	SigningKey string
```
Crevil, line 604: This field is actually no longer used, is it? So all the changes related to `ComitterDetails` could be reverted now that we just use `git` to lookup the data instead of parsing it in as command arguments?

kaspernissen:
You are absolutely right, we can simplify this a whole lot. I will do that 👍

cmd/server/http/http.go
kaspernissen: Yep. Fixed 👍

cmd/server/http/http.go
kaspernissen: Yes it does. Fixed 👍

kaspernissen:
I think all comments and suggestions have been dealt with 🌮

cmd/server/http/http_test.go
Crevil: You didn't push this?

kaspernissen:
My bad 🤦 Let's try one more time 🙏

PR #173, opened 2020-05-01
Add signing of commits from release-manager server

kaspernissen:
This PR adds functionality to sign commits in the release-manager server.

It's implemented the same way that flux supports, this, the gpg function is currently borrowed as I had problems with the import failing.


cmd/server/command/start.go
```diff
@@ -131,11 +133,29 @@ func NewStart(startOptions *startOptions) *cobra.Command {
...
+			// Import GPG Keys
+			if startOptions.gitConfigOpts.SigningKey != "" {
+				if len(startOptions.gitConfigOpts.GPGImportPaths) < 1 {
+					return errors.New("gpg signing key provided, but no import paths specified")
+				}
+				for _, p := range startOptions.gitConfigOpts.GPGImportPaths {
+					// lets just use flux' implementation on how to load keys
+					keyfiles, err := gpg.ImportKeys(p, false)
+					if err != nil {
+						return errors.New(fmt.Sprintf("failed to import GPG key(s) from %s", p))
```
Crevil, line 145: We can use `fmt.Errorf` for formatted errors.
```suggestion
						return fmt.Errorf("failed to import GPG key(s) from %s", p)
```

cmd/server/command/start.go
```diff
@@ -131,11 +133,29 @@ func NewStart(startOptions *startOptions) *cobra.Command {
...
+					}
+					if keyfiles != nil {
+						log.Infof("imported GPG key(s) from %s files %v", p, keyfiles)
```
Crevil, line 148: I don't understand this log line.
It will be something like `imported GPG key(s) from /a/path files [a.key b.key]`?

cmd/server/gpg/gpg.go
```diff
@@ -0,0 +1,87 @@
+// THIS IS BORROWED FROM FLUX!
```
Crevil, line 1: Can we add a reference to Flux's implementation? It's apache 2.0 licenced, so there is nothing wrong with using it.
It's just nice to have a link back to the original.

internal/git/git.go
```diff
@@ -32,10 +32,18 @@ var (
 	ErrUnknownGit         = errors.New("unknown git error")
 )
 
+type GitConfig struct {
+	User           string
+	Email          string
+	SigningKey     string
+	GPGImportPaths []string
```
Crevil, line 39: This field is not used in package `git` is it? Maybe we can move it out as a regular slice in `start.go` where it is used to load the keys?

cmd/server/command/start.go
kaspernissen: It just indicates that gpg keys from the specified paths have been imported and output their filenames.

internal/git/git.go
kaspernissen: Yes, that makes sense. I will fix this

PR #174, opened 2020-05-06
Add more logging on Flux notify slack

kaspernissen:


internal/flow/notify.go
```diff
@@ -71,7 +72,7 @@ func (s *Service) NotifyFluxEvent(ctx context.Context, event *http.FluxNotifyReq
 		err = s.Slack.NotifyFluxEventProcessed(ctx, commitMessage.ArtifactID, commitMessage.Environment, email, commitMessage.Service)
 		span.Finish()
 		if err != nil {
-			return errors.WithMessage(err, "post flux event processed private message")
+			return errors.WithMessage(err, fmt.Sprintf("post flux event processed private message; artifact: %s, env: %s, email: %s, service: %s", commitMessage.ArtifactID, commitMessage.Environment, email, commitMessage.Service))
```
Crevil, line 75: You can use `errors.WithMessagef()` instead.
```suggestion
			return errors.WithMessagef(err, "post flux event processed private message; artifact: %s, env: %s, email: %s, service: %s", commitMessage.ArtifactID, commitMessage.Environment, email, commitMessage.Service)
```

internal/flow/notify.go
```diff
@@ -2,6 +2,7 @@ package flow
 
 import (
 	"context"
+	"fmt"
```
Crevil, line 5: See below.
```suggestion
```

PR #175, opened 2020-05-07
Fix parsing of commitMessage to notify author via slack

kaspernissen:
This PR fixes an issue introduced after adding additional information to the commit message in the config repo. https://github.com/lunarway/release-manager/pull/173

This updates the parsing of the commit message to extract the git author of the original commit. This also introduces the possibility to send a message to the releaser, if we want in the future.

### DIFF
--- a/internal/flow/notify.go
+++ b/internal/flow/notify.go
@@ -2,7 +2,6 @@ package flow
 
 import (
 	"context"
-	"fmt"
 	"regexp"
 
 	"github.com/lunarway/release-manager/internal/flux"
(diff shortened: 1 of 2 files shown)


internal/flow/notify.go
```diff
@@ -94,11 +96,12 @@ func parseCommitMessage(commitMessage string) (FluxReleaseMessage, error) {
 	if len(matches) < 1 {
 		return FluxReleaseMessage{}, errors.New("not enough matches")
 	}
-
+	fmt.Printf("matches: %v", matches)
```
Crevil, line 99: Debug? 🙈 Might be nice to keep though as a formatted log line?

kaspernissen: 🤦

PR #177, opened 2020-05-19
S3 bucket upload

emilingerslev:


internal/flow/flow.go
```diff
@@ -290,20 +296,84 @@ func PushArtifact(ctx context.Context, gitSvc *git.Service, artifactFileName, re
...
-func listFiles(path string) {
+// PushArtifactToReleaseManager pushes an artifact to the release manager
+func PushArtifactToReleaseManager(ctx context.Context, releaseManagerClient *httpinternal.Client, artifactFileName, resourceRoot string) (string, error) {
+	artifactSpecPath := path.Join(resourceRoot, artifactFileName)
+	artifactSpec, err := artifact.Get(artifactSpecPath)
+	if err != nil {
+		return "", errors.WithMessagef(err, "path '%s'", artifactSpecPath)
+	}
+
+	path, err := releaseManagerClient.URLWithQuery(fmt.Sprintf("artifacts/create"), url.Values{})
```
Crevil, line 307: As we don't use query parameters we can use `Client.URL` instead.

```suggestion
	path, err := releaseManagerClient.URL(fmt.Sprintf("artifacts/create"))
```

internal/flow/flow.go
```diff
@@ -290,20 +296,84 @@ func PushArtifact(ctx context.Context, gitSvc *git.Service, artifactFileName, re
...
+	}
+
+	resp := httpinternal.ArtifactUploadResponse{}
+	err = releaseManagerClient.Do(http.MethodPost, path, httpinternal.ArtifactUploadRequest{
+		Artifact: artifactSpec,
+	}, &resp)
+	if err != nil {
+		return "", errors.WithMessage(err, "create artifact request failed")
+	}
+	log.WithFields("artifactID", artifactSpec.ID, "uploadURL", resp.ArtifactUploadURL).Infof("artifact upload URL created for %s", artifactSpec.ID)
```
Crevil, line 319: It's not that critical (or obvious) but we log with the `fmt` package in methods used by the CLIs only. 
I do however think we should move away from that and configure the `log` package in CLI mode instead.

But just wanted to mention it here for awareness :)

internal/flow/flow.go
```diff
@@ -290,20 +296,84 @@ func PushArtifact(ctx context.Context, gitSvc *git.Service, artifactFileName, re
...
+	files := listFiles(resourceRoot)
+
+	zipContent, err := zipFiles(files)
+	if err != nil {
+		return "", errors.WithMessage(err, "zip artifact failed")
+	}
+	log.WithFields("artifactID", artifactSpec.ID, "artifactFiles", files).Infof("artifact zip created for %s", artifactSpec.ID)
+
+	err = uploadFile(resp.ArtifactUploadURL, zipContent, artifactSpec)
+
```
Crevil, line 330:
```suggestion
```

internal/flow/flow.go
```diff
@@ -350,3 +420,65 @@ func (s *Service) cleanCopy(ctx context.Context, src, dest string) error {
...
+		if err != nil {
+			return nil, err
+		}
+		_, err = f.Write(fileContent)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// Make sure to check the error on Close.
```
Crevil, line 442: What's the reasoning for adding this comment? I'm not sure I understand the intent :)

internal/flow/flow.go
```diff
@@ -350,3 +420,65 @@ func (s *Service) cleanCopy(ctx context.Context, src, dest string) error {
...
+	return buf.Bytes(), nil
+}
+
+func uploadFile(url string, fileContent []byte, artifactSpec artifact.Spec) error {
+	// TODO: MD5
+	//h := md5.New()
+	//md5s := base64.StdEncoding.EncodeToString
+
+	req, err := http.NewRequest(http.MethodPut, url, bytes.NewReader(fileContent))
+	//req, err := http.NewRequest(http.MethodPut, url, strings.NewReader(""))
```
Crevil, line 457:
```suggestion
```

internal/flow/flow.go
```diff
@@ -350,3 +420,65 @@ func (s *Service) cleanCopy(ctx context.Context, src, dest string) error {
...
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("failed upload file to %s with status code %v and request id %v and body %s", url, resp.StatusCode, resp.Header["X-Amz-Request-Id"], string(body))
+	}
```
Crevil, line 481: We could omit the reading of the body to only be done in case of errors :) 
```suggestion

	if resp.StatusCode != http.StatusOK {
		body, err := ioutil.ReadAll(resp.Body)
		if err != nil {
			log.Errorf("Failed to read body from non-OK http request: %v", err)
		}

		return fmt.Errorf("failed upload file to %s with status code %v and request id %v and body %s", url, resp.StatusCode, resp.Header["X-Amz-Request-Id"], string(body))
	}
```

internal/flow/flow.go
```diff
@@ -350,3 +420,65 @@ func (s *Service) cleanCopy(ctx context.Context, src, dest string) error {
...
+		return err
+	}
+	req.Header.Set("x-amz-meta-artifact-spec", jsonSpec)
+
+	//req.Header.Set("Content-MD5", md5s)
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return err
+	}
+
```
Crevil, line 473: We should make sure to close the body here.

```suggestion
	defer resp.Body.Close()
```

internal/http/types.go
```diff
@@ -203,3 +203,11 @@ type DescribeArtifactResponse struct {
 	Service   string          `json:"service,omitempty"`
 	Artifacts []artifact.Spec `json:"artifacts,omitempty"`
 }
+
+type ArtifactUploadRequest struct {
+	Artifact artifact.Spec `json:"artifact,omitempty"`
+}
+
+type ArtifactUploadResponse struct {
+	ArtifactUploadURL string `json:"artifactUploadURL,omitempty"`
```
Crevil, line 212:
```suggestion
	ArtifactUploadURL string `json:"artifactUploadUrl,omitempty"`
```

internal/s3storage/s3storage.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/lunarway/release-manager/internal/log"
+)
+
+type Service struct {
+	bucketName string
+	logger     *log.Logger
+	s3client   *s3.S3
+}
+
+func Initialize(logger *log.Logger) (*Service, error) {
```
Crevil, line 20: Could this be `New` instead of `Initialize`? As it returns the allocated struct eg. instead of initializing a global value.

internal/s3storage/s3storage.go
```diff
@@ -0,0 +1,84 @@
...
+	sess, err := session.NewSession(&aws.Config{
+		Region: aws.String("eu-west-1"),
```
Crevil, line 22: Would be nice if this was configurable along with the bucket name. But we can always add that in another PR. :)

internal/slack/slack.go
```diff
@@ -15,13 +15,16 @@ type Client struct {
 	client        *slack.Client
 	emailMappings map[string]string
 	muteOptions   MuteOptions
+	skip          bool
```
Crevil, line 18: This isn't used, right?

cmd/server/http/create_artifact.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"github.com/lunarway/release-manager/internal/artifact"
+	httpinternal "github.com/lunarway/release-manager/internal/http"
+	"github.com/lunarway/release-manager/internal/log"
+	opentracing "github.com/opentracing/opentracing-go"
+)
+
+func createArtifact(payload *payload, artifactWriteStorage ArtifactWriteStorage) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		// copy span from request context but ignore any deadlines on the request context
```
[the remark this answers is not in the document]
emilingerslev: This will only create a pre-signed URL, so nothing will really happen, right?

cmd/server/http/s3webhook.go
```diff
@@ -0,0 +1,22 @@
+package http
```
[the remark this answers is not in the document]
emilingerslev: alright 👍

internal/flow/flow.go
emilingerslev: of course 🤦‍♂️

internal/flow/flow.go
emilingerslev: ok, so we can let it stay? I've seen quite a bit of difference in output, surely because logs are done in internal packages, which are then used in the CLI.

internal/flow/flow.go
emilingerslev: Just plain plagiarism 🤦 
https://golang.org/src/archive/zip/example_test.go

internal/flow/flow.go
emilingerslev: 👍

internal/flow/flow.go
emilingerslev: 👍

internal/slack/slack.go
emilingerslev: yeah, a leftover. I extended MuteOptions instead 👍

internal/s3storage/s3storage.go
emilingerslev: Yeah, like the auth options as arguments too 😊 let's grab them all together

internal/s3storage/s3storage.go
emilingerslev: Yeah, just wanted to tell the world that it's actually doing stuff - that is creating a bucket in the cloud. You sure you like `New` better considering that?

cmd/server/http/create_artifact.go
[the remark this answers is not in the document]
Crevil: Just for the correctness sake? It's a trivial change.

cmd/server/http/create_artifact.go
```diff
@@ -0,0 +1,48 @@
...
+		ctx := opentracing.ContextWithSpan(context.Background(), opentracing.SpanFromContext(r.Context()))
```
Crevil, line 16:
```suggestion
		ctx := r.Context()
```

internal/flow/flow.go
Crevil: Yes yes, let's align that in another PR

internal/s3storage/s3storage.go
Crevil: I understand the intention. Another approach would be to separate it out into two calls. A `New` function and `Initialize` method.

But what ever you like :) We can keep it as is.

internal/flow/flow.go
```diff
@@ -1,8 +1,13 @@
 package flow
 
 import (
+	"archive/zip"
+	"bytes"
 	"context"
 	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
```
Crevil, line 10: According to the build, this import is not used 🙈 
```suggestion
```

internal/flow/flow.go
emilingerslev: 👍

PR #178, opened 2020-05-22
Refactor artifact storage into separate package

Crevil:
This got a bit out of hand, but there are no functional changes except a subtle
bug fix. :) 

It is a change to package `flow` to extract the artifact storage layer into an
interface. The storage is still Git for now, but the interface is not
implemented by package `git` instead. This will eventually be replaced with AWS
S3 blob storage and with the new `flow.Storage` interface it is clear what a
storage layer needs to provide.

This will also make it simpler to implement in-memory storages for testing the
flows without touching external mechanisms.

The heart of the change is as mentioned the `flow.Storage` interface.

```go
type Storage interface {
	ArtifactExists(ctx context.Context, artifactID string) (bool, error)

	ArtifactSpecification(ctx context.Context, service, artifactID string) (artifact.Spec, error)
	ArtifactPaths(ctx context.Context, service, environment, branch, artifactID string) (specPath, resourcesPath string, close func(context.Context), err error)

	LatestArtifactSpecification(ctx context.Context, service, branch string) (artifact.Spec, error)
	LatestArtifactPaths(ctx context.Context, service, environment, branch string) (specPath, resourcesPath string, close func(context.Context), err error)

	ArtifactSpecifications(ctx context.Context, service string, count int) ([]artifact.Spec, error)
}
```

It allows the `flow` package to check for the existence of artifacts along with
fetching specifications for a specific artifact of a `n` artifacts.

The `ArtifactPaths` and `LatestArtifactPaths` methods are the methods providing
access to the actual resources of an artifact. Currently these are paths to the
file system, but I think we should come up with something better later on. Maybe
it will become clear with the S3 storage implementation what makes sense.

### Bug fix in promote commit messages

During the change I found a bug in the commit messages of promote operations.
The are currently resulting in the commit message `[env/service] release by
bso@lunar.app`. Note the missing artifact ID. This was introduced in
b1c9fe82f11e11cd2abb1e084ead74dc2f0a8e9c (#140).

### DIFF
--- a/internal/flow/flow.go
+++ b/internal/flow/flow.go
@@ -229,7 +229,7 @@ func envSpec(root, artifactFileName, service, env, namespace string) (artifact.S
 func (s *Service) sourceSpec(ctx context.Context, service, env, namespace string) (artifact.Spec, error) {
 	switch env {
 	case "dev":
-		return s.Storage.GetLatestArtifactSpecification(ctx, storage.ArtifactLocation{
+		return s.Storage.LatestArtifactSpecification(ctx, storage.ArtifactLocation{
 			Branch:  "master",
 			Service: service,
 		})
--- a/internal/flow/flow.go
+++ b/internal/flow/flow.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/lunarway/release-manager/internal/artifact"
 	"github.com/lunarway/release-manager/internal/copy"
-	"github.com/lunarway/release-manager/internal/flow/storage"
 	"github.com/lunarway/release-manager/internal/git"
 	"github.com/lunarway/release-manager/internal/log"
 	"github.com/lunarway/release-manager/internal/slack"
@@ -34,7 +33,7 @@ type Service struct {
 	Git              *git.Service
 	Tracer           tracing.Tracer
 	CanRelease       func(ctx context.Context, svc, branch, env string) (bool, error)
-	Storage          storage.Storage
+	Storage          Storage
 
 	PublishPromote           func(context.Context, PromoteEvent) error
 	PublishRollback          func(context.Context, RollbackEvent) error
@@ -46,6 +45,18 @@ type Service struct {
 	// NotifyReleaseHook is triggered in a Go routine when a release is completed.
 	// The context.Context is cancelled if the originating flow call is cancelled.
 	NotifyReleaseHook func(ctx context.Context, options NotifyReleaseOptions)
+}
+
+type Storage interface {
+	ArtifactExists(ctx context.Context, artifactID string) (bool, error)
+
+	ArtifactSpecification(ctx context.Context, service, artifactID string) (artifact.Spec, error)
+	ArtifactPaths(ctx context.Context, service, environment, branch, artifactID string) (specPath, resourcesPath string, close func(context.Context), err error)
+
+	LatestArtifactSpecification(ctx context.Context, service, branch string) (artifact.Spec, error)
+	LatestArtifactPaths(ctx context.Context, service, environment, branch string) (specPath, resourcesPath string, close func(context.Context), err error)
+
+	ArtifactSpecifications(ctx context.Context, service string, count int) ([]artifact.Spec, error)
 }
 
 type NotifyReleaseOptions struct {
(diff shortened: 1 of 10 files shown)


Crevil:
I can't seem to get the last commits up to GitHub. Everything is pushed from my machine but it's not updating in here. 🤔

Crevil:
There we are :)

internal/flow/promote.go
```diff
@@ -165,33 +158,30 @@ func (p *PromoteEvent) Unmarshal(data []byte) error {
...
 		actor := p.Actor
-		logger.Debugf("internal/flow: Promote: release hash '%v'", hash)
-		err = s.Git.Checkout(ctx, sourceConfigRepoPath, hash)
+
+		var (
+			artifactSourcePath, sourcePath string
+			closeSource                    func(context.Context)
+			err                            error
+		)
+		// when promoting to dev we use should look for the artifact instead of
```
kaspernissen, line 174:
```suggestion
		// when promoting to dev we should look for the artifact instead of
```

internal/flow/flow.go
```diff
@@ -46,6 +47,18 @@ type Service struct {
 	NotifyReleaseHook func(ctx context.Context, options NotifyReleaseOptions)
 }
 
+type Storage interface {
```
emilingerslev, line 50: Could be nice with this split into more files, right?

internal/flow/flow.go
```diff
@@ -46,6 +47,18 @@ type Service struct {
...
+	ArtifactExists(ctx context.Context, artifactID string) (bool, error)
+
+	ArtifactSpecification(ctx context.Context, service, artifactID string) (artifact.Spec, error)
+	ArtifactPaths(ctx context.Context, service, environment, branch, artifactID string) (specPath, resourcesPath string, close func(context.Context), err error)
```
emilingerslev, line 54: Docs would help here, since I don't find the naming too obvious.
What does `Paths` methods do?

internal/flow/promote.go
```diff
@@ -165,33 +158,30 @@ func (p *PromoteEvent) Unmarshal(data []byte) error {
...
-		err = s.Git.Checkout(ctx, sourceConfigRepoPath, hash)
+
+		var (
+			artifactSourcePath, sourcePath string
+			closeSource                    func(context.Context)
+			err                            error
+		)
+		// when promoting to dev we should look for the artifact instead of
+		// release as the artifact have never been released.
+		if environment == "dev" {
```
emilingerslev, line 176: This is a bit dangerous if we were to make policies that skip environments.
Coming to think of it, isn't it the same as `hamctl release --env dev --branch master`?

internal/flow/promote.go
```diff
@@ -256,3 +244,81 @@ func (s *Service) ExecPromote(ctx context.Context, p PromoteEvent) error {
...
+	if err != nil {
+		closeSource(ctx)
+		return "", "", nil, errors.WithMessagef(err, "checkout release hash '%s'", hash)
+	}
+	artifactPath := srcPath(sourceConfigRepoPath, service, "master", s.ArtifactFileName)
+	resourcesPath := srcPath(sourceConfigRepoPath, service, "master", environment)
+	return artifactPath, resourcesPath, closeSource, nil
+}
+
+func (s *Service) releaseExists(ctx context.Context, artifactID string) (bool, error) {
```
emilingerslev, line 275: What does this do? 🤔

internal/flow/promote.go
Crevil: It is, and is a result of our old Git promote between branches implementation.
It is the same as now, but just moved into this code section.

internal/flow/flow.go
Crevil: I've added documentation to the interface methods for better understanding.

internal/flow/flow.go
Crevil: Moved into a separate file.

internal/flow/promote.go
Crevil: Check for the existence of a release of provided  artifact ID.

Crevil:
Travis is building the wrong commit 🤦

Crevil:
Yes it is. 🙌

PR #180, opened 2020-05-25
Validate environment in synchronous step of release flows

kaspernissen:
This fixes a bug where events are not being consumed because of an error with the environment. 

This PR adds validation upfront, to ensure we can process the event - and further informs the caller of an error in the given environment flag.


Crevil:
I'll merge these changes into the other PR when its in master 👍

kaspernissen:
The test seems to fail? Can you see what the problem is, @Crevil ? 😄

Crevil:
Fixed by waiting 🙈

PR #183, opened 2020-05-31
AWS S3 read storage

Crevil:
This PR adds a `flow.ArtifactReadStorage` implementation for S3 stored artifacts. An additonal fallback storage that will use S3 as primary and fallback to Git is also added.


cmd/artifact/Makefile
```diff
@@ -1,10 +1,12 @@
+ARTIFACT_ID=master-`date +"%Y%m%d"`-`date +"%H%M"`
```
emilingerslev, line 1: Would be nice to move these Make targets to be available from the main make file, but i guess that's a separate PR.

internal/fallbackstorage/artifactreadstorage_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/lunarway/release-manager/internal/flow"
+	"github.com/lunarway/release-manager/internal/log"
+	"github.com/lunarway/release-manager/internal/tracing"
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/zap/zapcore"
+)
+
+var _ flow.ArtifactReadStorage = &fallbackstorage.Fallback{}
+
+func TestFallback_ArtifactSpecifications(t *testing.T) {
```
emilingerslev, line 18: Good work with the fallback tests @Crevil 💪

cmd/artifact/Makefile
Crevil: Yeah, agree. That would be great :)

PR #184, opened 2020-06-02
S3 subscription

emilingerslev:
This PR is kind of two-fold.

The first part is about the S3 Storage that gets support for registering an AWS SQS Queue automatically including:
- Creating the SQS Queue if not existing
- Defining permissions on that SQS Queue to let S3 Notification Service send messages to that queue
- Updating the S3 buckets Notification configuration to send all CreateObject notifications to that queue.
- Continuously listening for new events on the SQS queue while the release manager is running.

The second part is about refactoring the execution of Auto Release Policies out of the Github Webhook logic. It is now set up so that Github Webhooks trigger a new broker event type called `NewArtifact`, which then asynchronously triggers the same Auto Release Policy logic as before.
This is done so the same logic and the same events can also be used by S3 events. 

This consequently means that S3 events are being moved as quickly as possible from the external AWS Queue system to our main queue system, which makes the system easier to reason about

### DIFF
--- a/cmd/server/command/start.go
+++ b/cmd/server/command/start.go
@@ -353,7 +353,7 @@ func NewStart(startOptions *startOptions) *cobra.Command {
 				done <- errors.WithMessage(err, "broker")
 			}()
 
-			sqsHandler := func() error {
+			sqsHandler := func(msg string) error {
 				return nil
 			}
 
--- a/cmd/server/command/start.go
+++ b/cmd/server/command/start.go
@@ -413,7 +413,6 @@ func NewStart(startOptions *startOptions) *cobra.Command {
 					err := s3storageSvc.Close()
 					if err != nil {
 						log.Errorf("Failed to close s3 storage: %v", err)
-						done <- errors.WithMessage(err, "s3 storage")
 					}
 				}()
 			}
--- a/cmd/server/command/start.go
+++ b/cmd/server/command/start.go
@@ -164,14 +164,6 @@ func NewStart(startOptions *startOptions) *cobra.Command {
 			if err != nil {
 				return err
 			}
-			err = s3storageSvc.InitializeSQS()
-			if err != nil {
-				return err
-			}
-			err = s3storageSvc.InitializeBucket()
-			if err != nil {
-				return err
-			}
 
 			github := github.Service{
 				Token: *startOptions.githubAPIToken,
@@ -359,6 +351,26 @@ func NewStart(startOptions *startOptions) *cobra.Command {
 			go func() {
 				err := brokerImpl.StartConsumer(eventHandlers, errorHandler)
 				done <- errors.WithMessage(err, "broker")
+			}()
+
+			sqsHandler := func() error {
+				return nil
+			}
+
+			err = s3storageSvc.InitializeBucket()
+			if err != nil {
+				return err
+			}
+			err = s3storageSvc.InitializeSQS(sqsHandler)
+			if err != nil {
+				return err
+			}
+			defer func() {
+				err := s3storageSvc.Close()
+				if err != nil {
+					log.Errorf("Failed to close s3 storage: %v", err)
+					done <- errors.WithMessage(err, "s3 storage")
+				}
 			}()
 
 			sigs := make(chan os.Signal, 1)
--- a/cmd/server/command/start.go
+++ b/cmd/server/command/start.go
@@ -383,7 +383,10 @@ func NewStart(startOptions *startOptions) *cobra.Command {
 			if s3storageSvc != nil {
 				sqsHandler := func(msg string) error {
 					var s3event s3storage.S3Event
-					s3event.Unmarshal([]byte(msg))
+					err := s3event.Unmarshal([]byte(msg))
+					if err != nil {
+						return errors.WithMessage(err, "unmarshal S3Event")
+					}
 
 					if len(s3event.Records) == 0 {
 						log.With("sqsMessage", msg).Infof("Skipping SQS message because it does not look like S3 notification")
--- a/cmd/server/command/start.go
+++ b/cmd/server/command/start.go
@@ -359,6 +359,9 @@ func NewStart(startOptions *startOptions) *cobra.Command {
 			flowSvc.PublishRollback = func(ctx context.Context, event flow.RollbackEvent) error {
 				return brokerImpl.Publish(ctx, &event)
 			}
+			flowSvc.PublishNewArtifact = func(ctx context.Context, event flow.NewArtifactEvent) error {
+				return brokerImpl.Publish(ctx, &event)
+			}
 			defer func() {
 				err := brokerImpl.Close()
 				if err != nil {
@@ -388,7 +391,12 @@ func NewStart(startOptions *startOptions) *cobra.Command {
 					}
 
 					for _, record := range s3event.Records {
-						log.Infof("Got S3 event: %+v", record)
+						parts := strings.Split(record.S3.Object.Key, "/")
+						if len(parts) != 2 {
+							log.With("s3event", s3event).Infof("Got s3 object creation event on %s which can't be parsed to service and artifact id", record.S3.Object.Key)
+							continue
+						}
+						flowSvc.NewArtifact(ctx, parts[0], parts[1])
 					}
 					return nil
 				}
--- a/cmd/server/command/start.go
+++ b/cmd/server/command/start.go
@@ -203,6 +203,7 @@ func NewStart(startOptions *startOptions) *cobra.Command {
 				Git:              &gitSvc,
 				CanRelease:       policySvc.CanRelease,
 				Storage:          storage,
+				Policy:           &policySvc,
 				Tracer:           tracer,
 				// TODO: figure out a better way of splitting the consumer and publisher
 				// to avoid this chicken and egg issue. It is not a real problem as the
@@ -212,6 +213,7 @@ func NewStart(startOptions *startOptions) *cobra.Command {
 				PublishReleaseArtifactID: nil,
 				PublishReleaseBranch:     nil,
 				PublishRollback:          nil,
+				PublishNewArtifact:       nil,
 				// retries for comitting changes into config repo
 				// can be required for racing writes
 				MaxRetries: 3,
@@ -316,6 +318,14 @@ func NewStart(startOptions *startOptions) *cobra.Command {
 						return errors.WithMessage(err, "unmarshal event")
 					}
 					return flowSvc.ExecRollback(context.Background(), event)
+				},
+				flow.NewArtifactEvent{}.Type(): func(d []byte) error {
+					var event flow.NewArtifactEvent
+					err := event.Unmarshal(d)
+					if err != nil {
+						return errors.WithMessage(err, "unmarshal event")
+					}
+					return flowSvc.ExecNewArtifact(context.Background(), event)
 				},
 			}
 
--- a/cmd/server/command/start.go
+++ b/cmd/server/command/start.go
@@ -91,6 +91,10 @@ type configRepoOptions struct {
 	SSHPrivateKeyPath string
 }
 
+type s3storageOptions struct {
+	S3BucketName string
+}
+
 type startOptions struct {
 	slackAuthToken            *string
 	githubAPIToken            *string
@@ -99,6 +103,7 @@ type startOptions struct {
 	gitConfigOpts             *git.GitConfig
 	http                      *http.Options
 	broker                    *brokerOptions
+	s3storage                 *s3storageOptions
 	slackMutes                *slack.MuteOptions
 	gpgKeyPaths               *[]string
 	userMappings              *map[string]string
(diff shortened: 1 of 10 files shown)


internal/s3storage/s3storage.go
```diff
@@ -68,3 +163,60 @@ func (s *Service) CreateArtifact(artifactSpec artifact.Spec, md5 string) (string
...
+	return nil
+}
+
+func (s *Service) startSQSHandler(handler func(msg string) error) {
+	log.Infof("starting SQS handler")
+	s.sqsHandlerQuitChannel = make(chan struct{})
+	s.sqsHandlerErrorChannel = make(chan error, 1)
+	go func() {
+		for {
+			_, ok := <-s.sqsHandlerQuitChannel
```
Crevil, line 184: The initial check on the `sqsHandlerQuitChannel` handler is only checked on startup. It should instead be in a select statement with `output.Messages` as to continually validate whether the service is being closed.

```golang
output, err := s.sqsClient.ReceiveMessage(&sqs.ReceiveMessageInput{
if err != nil {
	// handle
}
for {
	select {
		case <-s.sqsHandlerQuitChannel:
			return
		case message := <-output.Messages:
			// handle
	}
}
```

emilingerslev: Why is it only checked on startup? `s.sqsClient.ReceiveMessage(&sqs.ReceiveMessageInput` is executed every 5 seconds, which is then followed by `_, ok := <-s.sqsHandlerQuitChannel`.

emilingerslev: Rewrote this, so lets see if you still think its a problem :)

cmd/server/http/http.go
```diff
@@ -739,23 +698,34 @@ func convertTimeToEpoch(t time.Time) int64 {
...
+var extractInfoFromCommitRegexNamesLookup = make(map[string]int)
+
+func init() {
+	for index, name := range extractInfoFromCommitRegex.SubexpNames() {
+		if name != "" {
+			extractInfoFromCommitRegexNamesLookup[name] = index
+		}
+	}
+}
+
```
Crevil, line 706: This is clever 👏 
Could we move it into the scope of `extractInfoFromCommit` where the `pattern` was defined before? Just to avoid global scoped logic?

cmd/server/command/start.go
```diff
@@ -357,6 +380,44 @@ func NewStart(startOptions *startOptions) *cobra.Command {
 				done <- errors.WithMessage(err, "broker")
 			}()
 
+			if s3storageSvc != nil {
+				sqsHandler := func(msg string) error {
+					var s3event s3storage.S3Event
+					s3event.Unmarshal([]byte(msg))
```
Crevil, line 386: The error is not checked here? `Unmarhsal` might fail if the payload for some reason isn't what we expect.
```suggestion
					err := s3event.Unmarshal([]byte(msg))
					if err != nil {
						return errors.WithMessage(err, "unmarshal S3Event")
					}
```

internal/s3storage/s3events.go
```diff
@@ -0,0 +1,52 @@
...
+	EventTime         string                `json:"eventTime,omitempty"`
+	EventName         string                `json:"eventName,omitempty"`
+	UserIdentity      S3EventRecordIdentity `json:"userIdentity,omitempty"`
+	RequestParameters map[string]string     `json:"requestParameters,omitempty"`
+	ResponseElements  map[string]string     `json:"responseElements,omitempty"`
+	S3                S3EventRecordData     `json:"s3,omitempty"`
+}
+
+type S3EventRecordIdentity struct {
+	PrincipalId string `json:"principalId,omitempty"` //"AWS:AROAIVBPOFACJXCN6JNTG:eki@lunar.app"
```
Crevil, line 30: There are multiple trailing comments in the structs that smell like testing output. Can we make them either more general if they need to be here, ie. remove Lunar and EKI, or can they be removed entirely?

internal/s3storage/s3storage.go
```diff
@@ -1,54 +1,149 @@
...
+	}
+	s.sqsQueueARN = *queueAttributes.Attributes[sqs.QueueAttributeNameQueueArn]
+	log.Infof("SQS queue arn acquired: %s", s.sqsQueueARN)
+
+	policy := `{
+		"Version": "2012-10-17",
+		"Id": "` + s.sqsQueueARN + `/SQSDefaultPolicy",
+		"Statement": [
+		 {
+			"Sid": "` + s.bucketName + `-s3-permission",
```
Crevil, line 85: Just a sec note. `s.bucketName` is inserted here directly from a flag and thus can be used to escape this statement eg bucket name `my-bucket" }]}/*`

It might not be any concern of ours as this is a server startup flag and not dynamically setup.

internal/s3storage/s3storage.go
```diff
@@ -1,54 +1,149 @@
...
+	 }`
+
+	_, err = s.sqsClient.SetQueueAttributes(&sqs.SetQueueAttributesInput{
+		QueueUrl: &s.sqsQueueURL,
+		Attributes: aws.StringMap(map[string]string{
+			sqs.QueueAttributeNamePolicy: policy,
+		}),
+	})
+	if err != nil {
+		log.With("policy", policy, "bucketName", s.bucketName, "Error", fmt.Sprintf("%s", err)).Errorf("Failed update policy: %s", err)
```
Crevil, line 108: The log package will serialize errors correctly, so no need for `fmt.Sprintf`.
```suggestion
		log.With("policy", policy, "bucketName", s.bucketName, "error", err).Errorf("Failed update policy: %v", err)
```

internal/s3storage/s3storage.go
```diff
@@ -68,3 +163,65 @@ func (s *Service) CreateArtifact(artifactSpec artifact.Spec, md5 string) (string
 
 	return uploadURL, nil
 }
+
+func (s *Service) Close() error {
```
Crevil, line 167: The method will panic if called multiple times as we cannot `close` a channel more than once.

It adds event more complexity to handle this, so maybe we should just add a note in the docs for now?
```suggestion
// Close closes the S3 storage service. Multiple calls to this method will result in a panic.
func (s *Service) Close() error {
```

internal/s3storage/s3storage.go
```diff
@@ -68,3 +163,65 @@ func (s *Service) CreateArtifact(artifactSpec artifact.Spec, md5 string) (string
...
+		for {
+			select {
+			case _, ok := <-s.sqsHandlerQuitChannel:
+				if !ok {
+					s.sqsHandlerErrorChannel <- nil
+					return
+				}
+			default:
+				// continue
+			}
```
Crevil, line 192: Could this not just be an `if` statement?

```suggestion
			_, ok := <-s.sqsHandlerQuitChannel:
			if !ok {
				s.sqsHandlerErrorChannel <- nil
				return
			}
```

internal/s3storage/s3storage.go
```diff
@@ -68,3 +163,65 @@ func (s *Service) CreateArtifact(artifactSpec artifact.Spec, md5 string) (string
...
+
+			if err != nil {
+				log.Errorf("failed to fetch sqs messages %v", err)
+				continue
+			}
+
+			for _, message := range output.Messages {
+				err := handler(*message.Body)
+				if err != nil {
+					log.With("messageID", *message.MessageId, "messageBody", *message.Body).Errorf("Failed handling SQS message. Error: %s", err)
```
Crevil, line 208: Is it intentional that we drop messages that could not be handled? A missing `continue` here results in `DeleteMessage` being called on every message.

cmd/server/command/start.go
```diff
@@ -357,6 +380,44 @@ func NewStart(startOptions *startOptions) *cobra.Command {
...
+
+					if len(s3event.Records) == 0 {
+						log.With("sqsMessage", msg).Infof("Skipping SQS message because it does not look like S3 notification")
```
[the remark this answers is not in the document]
emilingerslev: The output from AWS SDK is actually a `*string` (opposed to what you'd expect), so we get a string output 👍

cmd/server/command/start.go
```diff
@@ -357,6 +380,44 @@ func NewStart(startOptions *startOptions) *cobra.Command {
...
+				}
+				err = s3storageSvc.InitializeSQS(sqsHandler)
+				if err != nil {
+					return err
+				}
+				defer func() {
+					err := s3storageSvc.Close()
+					if err != nil {
+						log.Errorf("Failed to close s3 storage: %v", err)
+						done <- errors.WithMessage(err, "s3 storage")
```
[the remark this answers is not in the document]
emilingerslev: Makes sense 👍

internal/s3storage/s3storage.go
emilingerslev: Yeah, don't think its too much of a concern, due to running release-manager is going through another review process in `lunar-way-release-manager`.
Also if you tried to use tried to use `my-bucket" }]}/*` I would guess the CreateBucket step before will fail hard 😅

internal/s3storage/s3storage.go
emilingerslev: Oh, great! I stole that from https://github.com/lunarway/release-manager/blob/s3-notifications/internal/broker/amqp/consumer.go#L106:L106
but I see my mistake now 👍

internal/s3storage/s3storage.go
emilingerslev: Would rather it didn't blow up, but yeah, maybe the comment is good enough in this case.

internal/s3storage/s3storage.go
emilingerslev: I found that the `select` does not block, but `_, ok := <-s.sqsHandlerQuitChannel:` does block. Thats why it looks like it does.

internal/s3storage/s3storage.go
emilingerslev: Yeah, it was actually intentional, but I'm unsure if its a good idea. I'll add the `continue`

internal/s3storage/s3storage.go
Crevil: True. Adding a `closed` state that is atomically updated is the usual way, but to keep it simple enough in this controlled API I think we can leave at as is.

internal/s3storage/s3storage.go
Crevil: Yeah, good point. My mistake 👍

cmd/server/http/http.go
emilingerslev: missed a few first time around, including this. Yeah will look at that

cmd/server/http/http.go
```diff
@@ -472,50 +471,10 @@ func githubWebhook(payload *payload, flowSvc *flow.Service, policySvc *policyint
...
-					}
-				}
-				logger.Infof("http: github webhook: service '%s': auto-release from policy '%s' of %s to %s", commitInfo.Service, autoRelease.ID, releaseID, autoRelease.Environment)
-			}
-			if errs != nil {
-				logger.Errorf("http: github webhook: service '%s' branch '%s': auto-release failed with one or more errors: %v", commitInfo.Service, branch, errs)
-				unknownError(w)
-				return
+				logger.Infof("http: github webhook: service '%s': could not publish new artifact event for %s: %s", commitInfo.Service, commitInfo.ArtifactID, err)
+				w.WriteHeader(http.StatusInternalServerError)
```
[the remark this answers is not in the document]
Crevil: Looks like you missed the `return` addition here?

internal/s3storage/s3events.go
emilingerslev: Its and old leftover comment. Let's remove it 👍

cmd/server/command/start.go
emilingerslev: oops. Good catch 👍

Crevil:
There are still some unresolved comments. Care to take a look before I proceed with a new review?

cmd/server/http/http.go
emilingerslev: Oh, I see what you mean now. Yeah, we could do that, but this will run it as early as possible, this failing faster. Also it will only compile the regex once :)

internal/s3storage/s3storage.go
emilingerslev: Ok 👍

emilingerslev:
> There are still some unresolved comments. Care to take a look before I proceed with a new review?

Yeap, missed those with "more hidden comments" feature of github 🙄  Sorry!

cmd/server/http/http.go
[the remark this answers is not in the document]
emilingerslev: damn another missed one

cmd/server/http/http.go
Crevil: So will it in the handler setup. It is at startup where the server mux is created. 
Whether the error happens on importing `cmd/server/http` or using the handler function that requires the regex, I'd prefer the latter. But your call :)

cmd/server/http/http.go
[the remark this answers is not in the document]
emilingerslev: Fixed!

cmd/server/http/http.go
emilingerslev: True 💪

emilingerslev: Done 👍

PR #185, opened 2020-06-02
Ack and slack notify

emilingerslev:
This PR introduces an Error handler to the broker implementation, that supports a generic error handling when something goes wrong with receiving messages.
It also ack's all messages afterwards, so nothing gets blocked.

### DIFF
--- a/internal/broker/amqp/consumer.go
+++ b/internal/broker/amqp/consumer.go
@@ -60,7 +60,7 @@ func (c *consumer) Close() error {
 	return nil
 }
 
-func (c *consumer) Start(logger *log.Logger, handlers map[string]func([]byte) error) error {
+func (c *consumer) Start(logger *log.Logger, handlers map[string]func([]byte) error, errorHandler func(msgType string, msgBody []byte, err error)) error {
 	msgs, err := c.channel.Consume(
 		c.queue.Name,      // queue
 		"release-manager", // consumer
@@ -104,13 +104,15 @@ func (c *consumer) Start(logger *log.Logger, handlers map[string]func([]byte) er
 				"status":       "failed",
 				"responseTime": duration,
 				"error":        fmt.Sprintf("%+v", err),
-			}).Errorf("[consumer] [FAILED] Failed to handle message: nacking and requeing: %v", err)
+			}).Errorf("[consumer] [FAILED] Failed to handle message: trigger error handling and acking: %v", err)
 			// TODO: remove comments to allow for redelivery. This will put events
 			// into the unacknowledged state
 
 			// err := msg.Nack(false, true) if err != nil {
 			//  logger.WithFields("error", fmt.Sprintf("%+v", err)).Errorf("Failed to nack message: %v", err)
 			// }
+			errorHandler(msg.Type, msg.Body, err)
+			msg.Ack(false)
 			continue
 		}
 		logger.With("res", map[string]interface{}{
--- a/internal/broker/amqp/consumer.go
+++ b/internal/broker/amqp/consumer.go
@@ -104,7 +104,7 @@ func (c *consumer) Start(logger *log.Logger, handlers map[string]func([]byte) er
 				"status":       "failed",
 				"responseTime": duration,
 				"error":        fmt.Sprintf("%+v", err),
-			}).Errorf("[consumer] [FAILED] Failed to handle message: trigger error handling and acking: %v", err)
+			}).Errorf("[consumer] [FAILED] Failed to handle message: trigger error handling and nacking with no requeue: %v", err)
 			// TODO: remove comments to allow for redelivery. This will put events
 			// into the unacknowledged state
 
@@ -112,7 +112,7 @@ func (c *consumer) Start(logger *log.Logger, handlers map[string]func([]byte) er
 			//  logger.WithFields("error", fmt.Sprintf("%+v", err)).Errorf("Failed to nack message: %v", err)
 			// }
 			errorHandler(msg.Type, msg.Body, err)
-			msg.Ack(false)
+			msg.Nack(false, false)
 			continue
 		}
 		logger.With("res", map[string]interface{}{
--- a/internal/broker/amqp/consumer.go
+++ b/internal/broker/amqp/consumer.go
@@ -100,24 +100,32 @@ func (c *consumer) Start(logger *log.Logger, handlers map[string]func([]byte) er
 		err := handler(msg.Body)
 		duration := time.Since(now).Milliseconds()
 		if err != nil {
-			logger.With("res", map[string]interface{}{
+			res := map[string]interface{}{
 				"status":       "failed",
 				"responseTime": duration,
 				"error":        fmt.Sprintf("%+v", err),
-			}).Errorf("[consumer] [FAILED] Failed to handle message: trigger error handling and nacking with no requeue: %v", err)
-			// TODO: remove comments to allow for redelivery. This will put events
-			// into the unacknowledged state
-
-			// err := msg.Nack(false, true) if err != nil {
-			//  logger.WithFields("error", fmt.Sprintf("%+v", err)).Errorf("Failed to nack message: %v", err)
-			// }
-			errorHandler(msg.Type, msg.Body, err)
-			msg.Nack(false, false)
+				"redelivered":  msg.Redelivered,
+			}
+			if msg.Redelivered {
+				logger.With("res", res).Errorf("[consumer] [FAILED] Failed to handle message: trigger error handling and nacking with no requeue")
+				errorHandler(msg.Type, msg.Body, err)
+				err = msg.Nack(false, false)
+				if err != nil {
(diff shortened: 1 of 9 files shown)


internal/slack/slack.go
```diff
@@ -368,3 +371,27 @@ func (c *Client) NotifyK8SPodErrorEvent(ctx context.Context, event *http.PodErro
...
+	if err != nil {
+		// If user id somehow couldn't be found, post the message to #squad-nasa
+		userID = "#squad-nasa"
+	}
+
+	asUser := slack.MsgOptionAsUser(true)
+	attachments := slack.MsgOptionAttachments(slack.Attachment{
+		Title:      fmt.Sprintf(":boom: Release Manager failed :x:"),
+		Color:      MsgColorRed,
+		Text:       fmt.Sprintf("Failed handling event in release manager for:\nService: %s\nEnvironment: %s\nBranch: %s\nNamespace: %s\nError: %s", service, environment, branch, namespace, err),
```
Crevil, line 389: I think we should add something actionable here. This will lead to questions like “what do I do now?”

Can we indicate what the operation was and thus message them something like “your release of x to y failed. Please try again”

internal/broker/amqp/consumer.go
```diff
@@ -104,13 +104,15 @@ func (c *consumer) Start(logger *log.Logger, handlers map[string]func([]byte) er
...
-			}).Errorf("[consumer] [FAILED] Failed to handle message: nacking and requeing: %v", err)
+			}).Errorf("[consumer] [FAILED] Failed to handle message: trigger error handling and acking: %v", err)
 			// TODO: remove comments to allow for redelivery. This will put events
 			// into the unacknowledged state
 
 			// err := msg.Nack(false, true) if err != nil {
 			//  logger.WithFields("error", fmt.Sprintf("%+v", err)).Errorf("Failed to nack message: %v", err)
 			// }
+			errorHandler(msg.Type, msg.Body, err)
+			msg.Ack(false)
```
Crevil, line 115: Actually I think we should Nack with requeue=false instead to keep AMQP semantics intact. It is the same outcome but lets us track these failure handlings in existing metrics

internal/broker/amqp/consumer.go
```diff
@@ -104,13 +104,15 @@ func (c *consumer) Start(logger *log.Logger, handlers map[string]func([]byte) er
 				"status":       "failed",
 				"responseTime": duration,
 				"error":        fmt.Sprintf("%+v", err),
-			}).Errorf("[consumer] [FAILED] Failed to handle message: nacking and requeing: %v", err)
+			}).Errorf("[consumer] [FAILED] Failed to handle message: trigger error handling and acking: %v", err)
 			// TODO: remove comments to allow for redelivery. This will put events
 			// into the unacknowledged state
 
 			// err := msg.Nack(false, true) if err != nil {
```
Crevil, line 111: This can be removed with the new handling

internal/broker/amqp/consumer.go
emilingerslev: Sounds good. Wasn't totally sure of the outcome of that nack, but that makes sense 👍

internal/broker/amqp/consumer.go
emilingerslev: Shouldn't we log it somewhere though? Maybe in the error handler?

internal/slack/slack.go
emilingerslev: We could do that, and I was working on it. What stopped me was the many different messages depending on the issue. I see that I also forgot the event type, so this is for all promote, release, etc. Maybe we can do a `generateSlackMessage(evtType, service, environment, branch, namespace string)` that can return some of the better cases and fallback to something more common?

emilingerslev:
So I just added some "smarter" messages to slack notifications. Do you think that could work @Crevil? With a bit more work of course.

internal/slack/slack.go
```diff
@@ -368,3 +371,36 @@ func (c *Client) NotifyK8SPodErrorEvent(ctx context.Context, event *http.PodErro
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func generateSlackMessage(msgType, service, environment, branch, namespace string, err error) string {
+	switch {
+	case msgType == "promote" && service != "" && environment != "" && branch != "":
+		return fmt.Sprintf("Failed promoting %s #%s in %s. Try promiting again.\nError: %s", service, branch, environment, err)
```
kaspernissen, line 402: We could add the command for them to copy paste and highlight as code, maybe? 

Try to run the command again: `hamctl promote ...`

internal/slack/slack.go
```diff
@@ -368,3 +371,27 @@ func (c *Client) NotifyK8SPodErrorEvent(ctx context.Context, event *http.PodErro
...
 }
+
+func (c *Client) NotifyReleaseManagerError(ctx context.Context, service, environment, branch, namespace, authorEmail string, err error) error {
+	if c.muteOptions.ReleaseManagerError {
+		return nil
+	}
+	userID, err := c.getIdByEmail(ctx, authorEmail)
+	if err != nil {
+		// If user id somehow couldn't be found, post the message to #squad-nasa
+		userID = "#squad-nasa"
```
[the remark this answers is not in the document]
emilingerslev: Weighed forth and bac between different solutions, but yeah we could log instead.

internal/slack/slack.go
emilingerslev: Definitely. What if the info is malformed though? Most likely there is something wrong with the input, since it broke in the first place, right? Otherwise we could just requeue.

internal/slack/slack.go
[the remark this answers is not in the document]
kaspernissen: We do that in the other Slack error cases. Also, we should try to minimize the hard coded Lunar stuff to make the transformation to a real OSS project easier in the future 🙂

internal/slack/slack.go
Crevil: Actually, that is a great point. It is not due to input as we have fixed that condition in the syncronous step. The errros we’ve seen since are memory limits and alike. Maybe we should just requeue?

internal/slack/slack.go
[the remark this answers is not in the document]
Crevil: Agree. We get an alert on error logs, so that should be enough to notify us.

Crevil:
It could work I think, but ser my comment about requeuing elsewhere.
Maybe that is in fact the more correct aproach.

internal/slack/slack.go
emilingerslev: If we requeue, we should only do it a few times, then fallback to above behavior. Do we have possible counter to increment?

emilingerslev: Hmm, we can use the redeliver property, to ensure it doesn't continue forever, but only try once. Or we can send it back: https://stackoverflow.com/questions/23158310/how-do-i-set-a-number-of-retry-attempts-in-rabbitmq

internal/broker/amqp/consumer.go
Crevil: Log what? I refer to the comments?

internal/slack/slack.go
Crevil: Yes, we can setup some limitations for it, but that gets more complicated.
Let's just stick with the `Nack` right now to keep it simple.

Crevil: In order to let the `msgType` propagate into this handler, we could use a custom `error` implementation. That way the handlers can decide what the message should be and we can avoid the generic event?

emilingerslev: That's definitely also a way to go @Crevil. Does that mean each error handler would have to implement some interface? And then fall back to a generic one? 

I implemented the redeliver once, with the `msg.Redelivered` prop, for retry once behavior.

internal/slack/slack.go
```diff
@@ -368,3 +371,37 @@ func (c *Client) NotifyK8SPodErrorEvent(ctx context.Context, event *http.PodErro
 	}
 	return err
 }
+
+func (c *Client) NotifyReleaseManagerError(ctx context.Context, msgType, service, environment, branch, namespace, actorEmail string, err error) error {
+	if c.muteOptions.ReleaseManagerError {
+		return nil
+	}
+	userID, err := c.getIdByEmail(ctx, actorEmail)
+	if err != nil {
```
kaspernissen, line 380: I think it's totally fine to just return the `err`, as we handle the user email lookup and error logging in `c.getIdByEmail(ctx, actorEmail)`

emilingerslev: that will instead result in an error here:
https://github.com/lunarway/release-manager/pull/185/files/400949c1666eb8fe2849362e6a5b4274f37d143c#diff-2481863f7833169b61254081a2d3aedcR319-R322

Is that better?

kaspernissen: That's up to you :)

PR #188, opened 2020-06-04
Improve release manager error slack messages

emilingerslev:


internal/slack/slack.go
```diff
@@ -400,7 +400,9 @@ func (c *Client) NotifyReleaseManagerError(ctx context.Context, msgType, service
 func generateSlackMessage(msgType, service, environment, branch, namespace string, err error) string {
 	switch {
 	case msgType == "promote" && service != "" && environment != "" && branch != "":
-		return fmt.Sprintf("Failed promoting %s #%s in %s. Try promiting again.\nError: %s", service, branch, environment, err)
+		return fmt.Sprintf("Failed promoting %s #%s in %s. Validate the options you used and try promoting again.\nError: %s", service, branch, environment, err)
+	case msgType == "release.branch" && service != "" && environment != "" && branch != "":
+		return fmt.Sprintf("Failed releasing %s #%s to %s. Validate the options you used and try releasing again.\nError: %s", service, branch, environment, err)
```
Crevil, line 405: I don't think we should have the `Validate the options` part. We expect that any invalid options should be catched in the sync step, so this is more likely a temporary issue that needs to be retried. What do you think?

```suggestion
		return fmt.Sprintf("Failed promoting %s #%s in %s. Please, try promoting again to see if that solves the problem.\nError: %s", service, branch, environment, err)
	case msgType == "release.branch" && service != "" && environment != "" && branch != "":
		return fmt.Sprintf("Failed releasing %s #%s to %s. Validate the options you used and try releasing again.\nError: %s", service, branch, environment, err)
```

Further more we might want to add the `artifact` release case as well. If I release with `hamctl release --artifact master-12345 --env prod` we will not match any of these cases :(

emilingerslev: Just wanted to add one more case, not considering too many cases taking too much of my time. But we can easily add the additional one. What event is that?

Also on the note of the message, I'd like until we know more, that we ask them to validate the args, since we are not 100% sure they are correct. Should be easy to do for them.

Crevil: 👍  All right.

It is `release.artifact` I think. But we can just keep it as is if you wan't to get this out of the way.

emilingerslev: I'll look into it in another PR later 👍

PR #190, opened 2020-06-04
Add server side input validation

Crevil:
Currently a number of HTTP endpoints does not validate inputs before calling
internal methods. This leads to unexpected results like releasing to an empty
environment. (fixed client side with d25d488ffc1262feb0d9e55a9682441bbf421b0e
(#189)).

This change set adds a Validate method to all HTTP request models that is called
in the HTTP handlers to ensure that data is present before proceeding.


Crevil:
In the CLI cobra handles it, so no, there is nothing to reuse from there.

Great 👍

PR #191, opened 2020-06-06
Introduce Service Filtering

emilingerslev:
The idea here is to be able to run Release Manager for a subset of services. This will allow us to roll out
release manager with 2 (or more) bulk heads, limiting the impact of deploying a release manager with buggy behavior.

Also we could have release-manager 1 deploy release-manager 2 and the other way around.


Crevil:
I'm closing this for now as we properly won't look further into it for the time being. I'll leave the branch as is just to keep the work.

PR #194, opened 2020-06-15
Split server http.go into multiple files

emilingerslev:


emilingerslev:
> Can we move the test of `extractInfoFromCommit` into `commit_info_test.go` as well?

100p 🤓

PR #195, opened 2020-06-15
Promote using release

emilingerslev:


cmd/hamctl/command/promote.go
```diff
@@ -21,43 +18,31 @@ func NewPromote(client *httpinternal.Client, service *string) *cobra.Command {
...
 	}
-	command.Flags().StringVarP(&environment, "env", "e", "", "Environment to promote to (required)")
-	completion.FlagAnnotation(command, "env", "__hamctl_get_environments")
-	// errors are skipped here as the only case they can occour are if thee flag
-	// does not exist on the command.
-	//nolint:errcheck
-	command.MarkFlagRequired("env")
-	command.Flags().StringVarP(&namespace, "namespace", "n", "", "Namespace the service is deployed to (defaults to env)")
-	completion.FlagAnnotation(command, "namespace", "__hamctl_get_namespaces")
+	command.Flags().StringVarP(&toEnvironment, "env", "", "", "Alias for '--to-env' (deprecated)")
```
Crevil, line 43: I strongly disagree with this change 😄 
I understand that it can be confusing whether it is the to or from environment and that knowledge has probably been tribal since the creation of the command.

I'd like to challenge whether this is generally misunderstood/hard to grasp and then take action when we know more. I also think that this is not directly affecting the goal of this PR relate to changing the flow? If that's the case can we move it into another branch and then discuss the change out of band with the rest of the changes?

cmd/server/http/release.go
```diff
@@ -29,70 +30,50 @@ func release(payload *payload, flowSvc *flow.Service) http.HandlerFunc {
 		logger = logger.WithFields(
 			"service", req.Service,
 			"req", req)
```
Crevil, line 32: Would be nice to record the intents here :D

cmd/server/http/release.go
```diff
@@ -29,70 +30,50 @@ func release(payload *payload, flowSvc *flow.Service) http.HandlerFunc {
...
 			switch errorCause(err) {
 			case flow.ErrReleaseProhibited:
-				logger.Infof("http: release: service '%s' environment '%s' branch '%s' artifact id '%s': release rejected: branch prohibited in environment: %v", req.Service, req.Environment, req.Branch, req.ArtifactID, err)
-				if req.Branch != "" {
-					httpinternal.Error(w, fmt.Sprintf("branch '%s' cannot be released to environment '%s' due to branch restriction policy", req.Branch, req.Environment), http.StatusBadRequest)
-				} else {
-					httpinternal.Error(w, fmt.Sprintf("artifact '%s' cannot be released to environment '%s' due to branch restriction policy", req.ArtifactID, req.Environment), http.StatusBadRequest)
-				}
+				logger.Infof("http: release: service '%s' environment '%s' artifact id '%s': release rejected: branch prohibited in environment: %v", req.Service, req.Environment, req.ArtifactID, err)
+				httpinternal.Error(w, fmt.Sprintf("artifact '%s' cannot be released to environment '%s' due to branch restriction policy", req.ArtifactID, req.Environment), http.StatusBadRequest)
```
Crevil, line 50: I think it might still be valuable to report a specific message to the user based on the intent?
The flow is to me an implementation detail that bubbles out into the error message:

```
$ hamctl release --branch feature --env prod
Error: artifact 'feature-sha-sha' cannot be released to environment 'prod' due to branch restriction policy
```

This can lead to questions like "What is feature-sha-sha?", "I didn't write feature-sha-sha in --branch. What is going on?" etc.

cmd/hamctl/command/promote.go
emilingerslev: Reverting this, so this can be discussed elsewhere 👍

cmd/server/http/release.go
emilingerslev: did my best to try to do this. Let me know what you think

cmd/server/http/release.go
emilingerslev: Indeed! 👍  Added

cmd/hamctl/command/actions/artifact_id.go
```diff
@@ -0,0 +1,58 @@
...
+		return "", err
+	}
+	err = client.Do(http.MethodGet, path, nil, &describeResp)
+	if err != nil {
+		return "", err
+	}
+
+	if len(describeResp.Artifacts) == 0 {
+		// TODO: write better message:
+		return "", fmt.Errorf("no artifacts found on master branch")
```
Crevil, line 54: How about something like this? I think that will suffice for now?

```suggestion
		return "", fmt.Errorf("no artifact found from branch '%s'", branch)
```

This will give this experience:

```
$ hamctl promote --env dev
Error: no artifact found from branch 'master'
```

cmd/hamctl/command/promote.go
```diff
@@ -21,43 +18,34 @@ func NewPromote(client *httpinternal.Client, service *string) *cobra.Command {
...
-			} else {
-				fmt.Printf("[✓] Promotion of %s from %s to %s initialized\n", resp.Tag, resp.FromEnvironment, resp.ToEnvironment)
-			}
 
-			return nil
+			return actions.ReleaseArtifactID(client, *service, toEnvironment, artifactID, intent.NewPromoteEnvironment(fromEnvironment))
 		},
 	}
-	command.Flags().StringVarP(&environment, "env", "e", "", "Environment to promote to (required)")
-	completion.FlagAnnotation(command, "env", "__hamctl_get_environments")
```
Crevil, line 55: This annotation should be kept and used for both `env` and `from-env`.

cmd/hamctl/command/promote.go
```diff
@@ -21,43 +18,34 @@ func NewPromote(client *httpinternal.Client, service *string) *cobra.Command {
...
-	// errors are skipped here as the only case they can occour are if thee flag
-	// does not exist on the command.
-	//nolint:errcheck
+	command.Flags().StringVarP(&toEnvironment, "env", "", "", "Environment to promote to (required)")
```
Crevil, line 45: The alias `e` is removed here. Is that intentional?
```suggestion
	command.Flags().StringVarP(&toEnvironment, "env", "e", "", "Environment to promote to (required)")
```

cmd/hamctl/command/promote.go
```diff
@@ -21,43 +18,34 @@ func NewPromote(client *httpinternal.Client, service *string) *cobra.Command {
...
+	command.Flags().StringVarP(&fromEnvironment, "from-env", "", "", "Environment to promote from")
+
 	command.MarkFlagRequired("env")
-	command.Flags().StringVarP(&namespace, "namespace", "n", "", "Namespace the service is deployed to (defaults to env)")
-	completion.FlagAnnotation(command, "namespace", "__hamctl_get_namespaces")
```
Crevil, line 61: We still have, and use, the `namespace` variable set by this flag in `actions.ArtifactIDFromEnvironment()`. Shouldn't this still be here?

cmd/hamctl/command/root.go
```diff
@@ -48,7 +48,6 @@ func NewCommand(version *string) (*cobra.Command, error) {
 			c.HelpFunc()(c, args)
 		},
 	}
-	command.AddCommand(NewPromote(&client, &service))
```
Crevil, line 51: The changes to `cmd/hamctl/command/promote.go` doesn't really make sense when we don't have the command anymore. Maybe we should just remove the file completely for now like on the server?

cmd/server/http/describe.go
```diff
@@ -168,3 +170,47 @@ func describeArtifact(ctx context.Context, payload *payload, flowSvc *flow.Servi
...
+func describeLatestArtifacts(ctx context.Context, payload *payload, flowSvc *flow.Service, service string) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if emptyString(service) {
+			requiredFieldError(w, "service")
+			return
+		}
+		values := r.URL.Query()
+		branch := values.Get("branch")
+		if emptyString(branch) {
+			branch = "master"
```
Crevil, line 183: Just an observation. This default value. When will we use the endpoint without? I'm thinking whether it should be required on the server and let the client default it? I dunno. 🤷‍♂️

internal/flow/release.go
```diff
@@ -104,117 +29,14 @@ func releaseConfigurationExists(resourcePath string) error {
...
-	Environment string `json:"environment,omitempty"`
-	Namespace   string `json:"namespace,omitempty"`
-	ArtifactID  string `json:"artifactID,omitempty"`
-	Branch      string `json:"branch,omitempty"`
-	Actor       Actor  `json:"actor,omitempty"`
+	Service     string        `json:"service,omitempty"`
+	Environment string        `json:"environment,omitempty"`
+	Namespace   string        `json:"namespace,omitempty"`
+	ArtifactID  string        `json:"artifactID,omitempty"`
+	Branch      string        `json:"branch,omitempty"`
```
Crevil, line 37: Is the `Branch` field still used now or is it just kept around for easy printing in the async flow?

internal/intent/intent.go
```diff
@@ -0,0 +1,86 @@
...
+const (
+	TypeReleaseArtifactID = "ReleaseArtifactID"
+	TypeReleaseBranch     = "ReleaseBranch"
+	TypePromote           = "Promote"
+	TypeRollback          = "Rollback"
+	TypeAutoRelease       = "AutoRelease"
+)
+
+type Intent struct {
+	Type          string              `json:"intentType,omitempty"`
```
Crevil, line 14: Is this deliberately using another JSON field name?

cmd/hamctl/command/promote.go
emilingerslev: oh yeah 👍  adding it for `--from-env` too

cmd/hamctl/command/promote.go
emilingerslev: No, lets keep it 👍

cmd/hamctl/command/promote.go
emilingerslev: hmm, you're right. Will have to remember this one 🤔

cmd/hamctl/command/root.go
emilingerslev: This was actually a mistake. Had readded this 🤔

cmd/server/http/describe.go
emilingerslev: sounds good with me 👍

internal/flow/release.go
emilingerslev: Thought we could remove it, but it is used by `storage.ArtifactPaths` 😔
I think we can remove it later on when we pension git artifact storage.

internal/intent/intent.go
emilingerslev: no... Fixing! 👍

emilingerslev:
Fixed all that was easy fixable 👍  The only thing I've waited with fixing, is the `Branch` in `ReleaseArtifactIDEvent`, since it is still used by storage 😔
I suggest we fix it when we remove git storage.

cmd/server/http/describe.go
```diff
@@ -180,7 +180,8 @@ func describeLatestArtifacts(ctx context.Context, payload *payload, flowSvc *flo
 		values := r.URL.Query()
 		branch := values.Get("branch")
 		if emptyString(branch) {
-			branch = "master"
+			httpinternal.Error(w, "missing required branch parameter", http.StatusBadRequest)
```
Crevil, line 183: This could be `requiredFieldError` instead?

```suggestion
			requiredFieldError(w, "branch")
```

cmd/hamctl/command/actions/artifact_id.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return "", err
+	}
+	err = client.Do(http.MethodGet, path, nil, &describeResp)
+	if err != nil {
+		return "", err
+	}
+
+	if len(describeResp.Artifacts) == 0 {
+		// TODO: write better message:
```
Crevil, line 53: We can remove this TODO, right? :D

cmd/server/http/describe.go
emilingerslev: Yes! Was looking for that 🙄 🤔 💪

cmd/hamctl/command/actions/artifact_id.go
emilingerslev: 💨

Crevil:
Happy bear 🐻

emilingerslev:
A happy 🐻  is a true thunder 🧸

emilingerslev:
oh, the bliss of a huge PR lifting from your responsibility load 😌

PR #203, opened 2020-07-02
Fix commit parsing issue

emilingerslev:
Some commits weren't parsed correctly, but this makes commit title parsing less important, and will always try to read fields first.

### DIFF
--- a/internal/commitinfo/commitinfo.go
+++ b/internal/commitinfo/commitinfo.go
@@ -68,11 +68,11 @@ func ParseCommitInfo(commitMessage string) (CommitInfo, error) {
 	}
 
 	matches := parseCommitInfoFromCommitMessageRegex.FindStringSubmatch(convInfo.Message)
-	if matches == nil {
-		return CommitInfo{}, errors.Wrap(ErrNoMatch, fmt.Sprintf("commit message '%s' does not match expected message structure", convInfo.Message))
+	if matches == nil && !convInfo.HasField(FieldReleaseIntent) {
+		return CommitInfo{}, errors.Wrap(ErrNoMatch, fmt.Sprintf("commit message '%s' does have a Release-intent field and did not match expected message structure", convInfo.Message))
 	}
 
-	if matches[parseCommitInfoFromCommitMessageRegexLookup.Type] == "artifact" {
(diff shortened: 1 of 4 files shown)


internal/commitinfo/commitinfo.go
```diff
@@ -68,11 +68,11 @@ func ParseCommitInfo(commitMessage string) (CommitInfo, error) {
 	}
 
 	matches := parseCommitInfoFromCommitMessageRegex.FindStringSubmatch(convInfo.Message)
-	if matches == nil {
-		return CommitInfo{}, errors.Wrap(ErrNoMatch, fmt.Sprintf("commit message '%s' does not match expected message structure", convInfo.Message))
+	if matches == nil && !convInfo.HasField(FieldReleaseIntent) {
+		return CommitInfo{}, errors.Wrap(ErrNoMatch, fmt.Sprintf("commit message '%s' does have a Release-intent field and did not match expected message structure", convInfo.Message))
```
Crevil, line 72: In this case id does **not** have a release intent, right?

internal/commitinfo/commitinfo.go
```diff
@@ -68,11 +68,11 @@ func ParseCommitInfo(commitMessage string) (CommitInfo, error) {
 	}
 
 	matches := parseCommitInfoFromCommitMessageRegex.FindStringSubmatch(convInfo.Message)
-	if matches == nil {
```
Crevil, line 71: Don't we still need a check on just `matches == nil` ? With the extra case we might get to the bottom of this method with `matches == nil`?

Same in the case that `matches != nil` but non of the `service`, `artifactID` etc. checks match, right?

internal/commitinfo/commitinfo.go
emilingerslev: Ah 🤦  Fixing!

internal/commitinfo/commitinfo.go
emilingerslev: No, we don't care about the commit message parsing if there is a "Release-intent". We can read it all from fields, so there isn't a real reason for trying to parse the message, other than backward compatibility.

Crevil: Ok. Makes sense 👍

PR #204, opened 2020-07-02
Fix Flux Event commit parsing

emilingerslev:


emilingerslev:
@Crevil we keep getting errors on one of the builds periodically. Do we know why?

PR #213, opened 2020-08-21
Add amqpextra reconnecting broker implementation

Crevil:
Currently the amqp broker implementation looses connectivity if the RabbitMQ
instance it is connected to gets OOMKilled in kubernetes.

Instead of trying to fix the complex reconnection logic in package amqp the
module github.com/makasim/amqpextra is adds as a dependency. It transparently
handles reconnections correctly along with simplifying our own implementation by
orders of magnitude.

The behaviour is the same as package amqp and the broker has been replaced in
the server start command.

A utility CLI amqp is added used to simulate workloads on a RabbitMQ instance.
It is used to flood the instance while the server is connected to RabbitMQ and
ensure that we recover correctly after OOMKilled events.

### DIFF
--- a/cmd/amqp/main.go
+++ b/cmd/amqp/main.go
@@ -1,0 +1,74 @@
+package main
+
+import (
+	"flag"
+	"log"
+	"math/rand"
+	"sync"
+
+	"github.com/streadway/amqp"
+)
+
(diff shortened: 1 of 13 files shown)


kaspernissen (approved):
Nice 🚀
\lgtm

PR #215, opened 2020-11-27
feature/Release Manager Url Environment Variable

mahlunar:
These changes enable us to change the release manager url via environment variables instead of defaulting to the dev 
lunarway release manager

** Caution **

We need to make sure that the environment variable is set. Otherwise the cli will stop working

- Set http url default value based on os env variable
- Updated push artifact command to also use default os env variables

### DIFF
--- a/cmd/artifact/command/push.go
+++ b/cmd/artifact/command/push.go
@@ -3,7 +3,7 @@ package command
 import (
 	"context"
 	"fmt"
-	"os"
+	"golang.org/x/tools/go/ssa/interp/testdata/src/os"
 	"path"
 
 	"github.com/lunarway/release-manager/internal/flow"
@@ -48,8 +48,8 @@ func pushCommand(options *Options) *cobra.Command {
 			return nil
 		},
 	}
-	command.Flags().StringVar(&releaseManagerClient.BaseURL, "http-base-url", os.Getenv("HAMCTL_URL"), "address of the http release manager server")
-	command.Flags().StringVar(&releaseManagerClient.Metadata.AuthToken, "http-auth-token", os.Getenv("HAMCTL_AUTH_TOKEN"), "auth token for the http service")
+	command.Flags().StringVar(&releaseManagerClient.BaseURL, "http-base-url", os.Getenv("ARTIFACT_URL"), "address of the http release manager server")
+	command.Flags().StringVar(&releaseManagerClient.Metadata.AuthToken, "http-auth-token", "", "auth token for the http service")
 
 	// errors are skipped here as the only case they can occour are if thee flag
 	// does not exist on the command.
--- a/cmd/artifact/command/push.go
+++ b/cmd/artifact/command/push.go
@@ -3,6 +3,7 @@ package command
 import (
 	"context"
 	"fmt"
+	"os"
 	"path"
 
 	"github.com/lunarway/release-manager/internal/flow"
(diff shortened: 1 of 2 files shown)


cmd/artifact/command/push.go
```diff
@@ -47,8 +48,8 @@ func pushCommand(options *Options) *cobra.Command {
 			return nil
 		},
 	}
-	command.Flags().StringVar(&releaseManagerClient.BaseURL, "http-base-url", "https://release-manager.dev.lunarway.com", "address of the http release manager server")
-	command.Flags().StringVar(&releaseManagerClient.Metadata.AuthToken, "http-auth-token", "", "auth token for the http service")
+	command.Flags().StringVar(&releaseManagerClient.BaseURL, "http-base-url", os.Getenv("HAMCTL_URL"), "address of the http release manager server")
+	command.Flags().StringVar(&releaseManagerClient.Metadata.AuthToken, "http-auth-token", os.Getenv("HAMCTL_AUTH_TOKEN"), "auth token for the http service")
```
Crevil, line 52: This will leak thr token to the console if you use --help.

I know it is done this way in hamct as well, but I'd rather we found a way to avoid it, and instead wrote `from HAMCTL_..` in the value.

cmd/artifact/command/push.go
```diff
@@ -47,8 +48,8 @@ func pushCommand(options *Options) *cobra.Command {
 			return nil
 		},
 	}
-	command.Flags().StringVar(&releaseManagerClient.BaseURL, "http-base-url", "https://release-manager.dev.lunarway.com", "address of the http release manager server")
-	command.Flags().StringVar(&releaseManagerClient.Metadata.AuthToken, "http-auth-token", "", "auth token for the http service")
+	command.Flags().StringVar(&releaseManagerClient.BaseURL, "http-base-url", os.Getenv("HAMCTL_URL"), "address of the http release manager server")
```
Crevil, line 51: Maybe we should use `ARTIFACT_*` instead as that's the name of the CLI?

Crevil (changes requested):

cmd/artifact/command/push.go
mahlunar: That makes sense. What about setting the value(s) based on the the arg or os.GetEnv functions and leave the default value empty here? I would expect arg to take precedence if it is set?

kaspernissen: I think leaving the default to be empty fine, then just let the arg take precedence, check the env var, and if none is set, break out?

cmd/artifact/command/push.go
kaspernissen: Yes, `HAMCTL` does not make that much sense. Either we use the cmd name, or just `SERVER_URL` or similar?

mahlunar: If we change the existing prefix then we need a plan for how to migrate the environment variables.

Crevil: This CLI does not use env variables today, does it? It is solely flag based (and actually I think that is OK TBH. It is for CI purposes only.

mahlunar: Makes sense. I confused the sub projects.

cmd/artifact/command/push.go
Crevil: Agree. There is a trick somewhere that just travers the flag set and checks if an env variable is available. That would be pretty nice so all flags can be set by env variables or a setting.

Pseudo coded like this:

```go
for _, flag := range command.Flags() {
  envVar := fmt.Sprintf("HAMCTL_%s",  strings.Replace(flag.Name(), "-", "_"))
  v, ok := os.LookupEnv(envVar)
  if !ok {
    continue
  }
  flag.Value(v)
}
```

The above makes no sense in practice regarding code. Just to show how it could work.

cmd/hamctl/command/root.go
```diff
@@ -34,6 +34,15 @@ func NewCommand(version *string) (*cobra.Command, error) {
...
 				return s.Vars.Service
 			})
+
+			if client.BaseURL == "" {
+				client.BaseURL = os.Getenv("HAMCTL_URL")
+			}
+
+			if client.Metadata.AuthToken == "" {
+				client.Metadata.AuthToken = os.Getenv("HAMCTL_AUTH_TOKEN")
+			}
```
mahlunar, line 44: I tested this part by running zplug update and restarting my terminal

kaspernissen (approved):
\lgtm

Crevil (approved):
LGTM

PR #216, opened 2020-11-30
Use native Git for reading configuration

Crevil:
Currently we use go-git to parse Git configuration, but the parsing is based on
static file lookups in a local .git/config file or the blobal .gitconfig. This
does not support Git's includeIf directives and also make hierarchical
configuraiton unusable by hamctl.

This change introduces use of the git CLI instead and thus delegates precedence
rules to native Git. This adds support for includeIf along with also making it
simpler to understand how it is done.

### DIFF
--- a/internal/git/config.go
+++ b/internal/git/config.go
@@ -1,0 +1,50 @@
+package git
+
+import (
+	"io/ioutil"
+	"os/exec"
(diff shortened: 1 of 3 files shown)


mahlunar (approved):
LGTM

PR #219, opened 2020-12-15
Remove github.com/weaveworks/flux dependency

Crevil:
Currently we depend on the github.com/weaveworks/flux module to parse flux
websocket events. We only rely on a couple of fields in the data model but
depend on all transitive dependencies of the flux project.

Further more the flux module has moved into github.com/fluxcd/flux after being
adopted by CNCF. A change in the new module structure does not allow us to
import the used packages anymore.

This change removes the dependency by copying the used models into this project.
It has little impact on readability but simplifies the dependency tree.

### DIFF
--- a/cmd/daemon/flux/exporter.go
+++ b/cmd/daemon/flux/exporter.go
@@ -5,20 +5,19 @@ import (
 	"fmt"
 	"net/http"
 
+	"github.com/lunarway/release-manager/internal/flux"
 	httpinternal "github.com/lunarway/release-manager/internal/http"
-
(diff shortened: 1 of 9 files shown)


mahlunar (approved):
lgtm

Crevil:
Just to add on top of the change description.
This also reduces the binary sizes of the different components notably.

```diff
- 19M artifact
+ 16M artifact
- 43M daemon
+ 39M daemon
- 18M hamctl
+ 15M hamctl
- 28M server
+ 27M server
```

PR #221, opened 2021-01-11
Check error from release action

Crevil:
Currently if the describe releases action fails when running hamctl describe
release no error is reported but an empty describe output is shown. This happens
as the err is not checked from the action call in hamctl.

```
$ hamctl describe release --service svc --env test --http-base-url http://nonsence
Service:
Environment:
```

This change adds an error check.

```
$ hamctl describe release --service svc --env test --http-base-url http://nonsence
Error: could not connect to the release-manager server. Are you connected to the internet and, if required, a VPN?
```


mahlunar (approved):
LGTM

PR #222, opened 2021-01-13
Fix missing releases in describe and rollback

Crevil:
Currently some releases are not reported when using `hamctl describe release` and
`hamctl rollback`.

This happens as on each matched release we checkout the repository on the found
hash and inspect some files. The `HEAD` is moved on each of these checkouts which
makes our search offset wrong.

This change fixes the bug by resetting the branch to the original `HEAD` after
each checkout ensuring our search offset is aligned.

It is not possible to test the `DescribeRelease` method in a simple way as it
depends on multiple git details and the file system. Because of this the change
set includes a unit test that only validates the core git logic of the flow in a
controlled manner. Eventually we should refactor the flows to be more testable
and move over to testing the actual implementation.

### DIFF
--- a/internal/flow/describe_release.go
+++ b/internal/flow/describe_release.go
@@ -48,6 +48,13 @@ func (s *Service) DescribeRelease(ctx context.Context, environment, service stri
 		return DescribeReleaseResponse{}, errors.WithMessagef(err, "clone into '%s'", sourceConfigRepoPath)
 	}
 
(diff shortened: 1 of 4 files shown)


mahlunar (approved):
LGTM

mahlunar:
Do you need to bump the version?

Crevil:
Yes, we will update that when the PRs are merged.

PR #224, opened 2021-01-22
Fix bug in server slack user lookup

Crevil:
Currently user-mappings on the server does not work. This is due to a bug
introduced in 73d80a895f35015e30975ef9b75fcaea14493926 where email suffix
validation was added to artifact.

The feature was not implemented correctly for the muteable slack client and thus
messages would be suppressed for non-slack workspace emails.

This change introduces the email-suffix flag to the server to let it know when
to lookup emails in the user mappings. To be able to test this the slack client
was changed to accept an interface of the upstream slack client instead of just
a token.


kaspernissen (approved):
Great catch!

\lgtm 🙌

PR #225, opened 2021-01-24
Extract version command setup

Crevil:
Currently the version commands are attached to the root commands unlike any
other of the commands. This is odd and unexpected.

This change moves the version commands into their own files and attaches them to
the root like every other command.

Furhter more the make targets for building the Go binaries are updated to
default to the current head git sha as version. This is mostly for testing but
also ensures that any distribution of a locally built binary is trackable.


kaspernissen (approved):
\lgtm

PR #226, opened 2021-02-04
[Spike] Enable platform env

mahlunar:
Spike for enabling a platform environment.

Looks like there is a lot of DRY cases in the code.

Could be awesome to specify environments in configuration to be customized.

### DIFF
--- a/internal/flow/flow.go
+++ b/internal/flow/flow.go
@@ -110,6 +110,7 @@ type StatusResponse struct {
 	Dev               Environment `json:"dev,omitempty"`
 	Staging           Environment `json:"staging,omitempty"`
 	Prod              Environment `json:"prod,omitempty"`
+	Platform          Environment `json:"platform,omitempty"`
 }
 
 type Actor struct {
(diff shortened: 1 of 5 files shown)


kaspernissen:
Yes, this should be specified in as an argument when running the server.

kaspernissen (approved):
Let's add this for now. As far as I know we already have this in the backlog somewhere 🙈

mahlunar:
> Let's add this for now. As far as I know we already have this in the backlog somewhere 🙈

Great. I have no idea if it works yet. I will refine it a bit and ask a review when ready.

internal/flow/flow.go
```diff
@@ -173,6 +174,21 @@ func (s *Service) Status(ctx context.Context, namespace, service string) (Status
 	}
 	defer span.Finish()
```
Crevil, line 175: This must be changed to a non-defer version as we do not return immediately after this call anymore. I can see that this is actually a bug in the previous lookups as well. Note in below trace how the `artifact spec for environment` all ends at the same time.

image


This can be fixed for all of them in another PR though.

Crevil (approved):
I think we should plan on removing this output and replace it by a more developer friendly view based on git as Emil has tested out.

My only concern with this change is that for all the developers not using the platform cluster, this new entry will show up 🤔

mahlunar:
> I think we should plan on removing this output and replace it by a more developer friendly view based on git as Emil has tested out.
> 
> My only concern with this change is that for all the developers not using the platform cluster, this new entry will show up 🤔

That is a valid concern. Let's say we at some point introduce more clusters then it will be even more noise.

mahlunar:
> > I think we should plan on removing this output and replace it by a more developer friendly view based on git as Emil has tested out.
> > My only concern with this change is that for all the developers not using the platform cluster, this new entry will show up 🤔
> 
> That is a valid concern. Let's say we at some point introduce more clusters then it will be even more noise.

We could create some configuration option so it only shows selected environments?

PR #227, opened 2021-02-15
Detect branch behind origin on git push

Crevil:
Currently if the release-manager gets behind the origin, eg. missing webhook, it
is not able to perform any push commands.

The master repo update mechanism is not triggered as we do not detect the error
case correctly. Currently we only detect if the origin has commits that are not
integrated locally that would conflict. Fast forward changes is thus ignored.

This change adds an additional error case to detect fast-forwardable errors and
report them as ErrBranchBehindOrigin. This should make the manager able to
recover from the state by updating the master branch and retrying the operation.

The implementation is changed to look like the `isKnownGitError` where we do a bit
comparision instead of using a regular expression. This is faster, cheaper and
simpler to read and understand.

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -8,7 +8,6 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
-	"regexp"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


kaspernissen (approved):
Lovely ❤️

PR #228, opened 2021-02-15
JSON syntax error as connectivity error in hamctl

Crevil:
Currently if hamctl cannot connect to the release-manager it will write an error
helping the client to understand this. If the connection is OK but a revers
proxy or alike does not grant access, eg. nginx IP allowlists, and returns an
HTML page we fail in parsing the response and the user is presented with a
non-descriptive error message.

    Error: response status 403 Forbidden: unmarshal error response: invalid character '<' looking for beginning of value

This change adds an error check on the decoding to detect this case and instead
report the connectivity error to the client.

    Error: could not connect to the release-manager server. Are you connected to the internet and, if required, a VPN?


kaspernissen (approved):
\lgtm

PR #229, opened 2021-02-15
Add support for multiple env in hamctl release

Crevil:
Currently the env flag of hamctl release only supports a single environment. It
is a common use case however to release to multiple environments in one go.

This change adds support for a comma separated list of environments. For each
found environment a release action is issued to the manager handling it with no
difference.

A change is necessary on the output of the release command as there can now be
partial failures. Each environment is now outputtet with its result.

    Release of service service-name using branch master
    [✓] Release of master-1-2 to dev initialized
    [✓] Release of master-1-2 to staging initialized

In case of failures it will look like this.

    Release of service service-name using branch master
    [X] cannot release master-1-2 to environment dev due to branch restriction policy (reference: GUID)
    [✓] Release of master-1-2 to staging initialized

As the error output is now line based, the reference ID is printet on the same
line as the error instead of on a new line.

### DIFF
--- a/cmd/hamctl/command/actions/release.go
+++ b/cmd/hamctl/command/actions/release.go
@@ -8,26 +8,49 @@ import (
 	"github.com/lunarway/release-manager/internal/intent"
 )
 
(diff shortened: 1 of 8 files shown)


kaspernissen (approved):
\lgtm 🚀

PR #232, opened 2021-02-16
Fork internal Lunar amqp library

Crevil:
This change introduces package internal/amqp which is a fork of Lunar's internal
AMQP transport. It handles most of the things the current amqpextra
implemenation does, but in a more robust way. It also has a number of features
that is currently not available today, eg. publisher confirms.

This should help us maintain the implementation in the future as we can follow
changes in the upstream in a closed package boundary.

Package amqpextra is updated to use this implementation instead which reduces
the complexity of that package significantly albeit most of it is practically
just moved into package amqp.


kaspernissen (approved):
\lgtm

PR #235, opened 2021-03-22
Restrict arguments to hamctl

Crevil:
Currently it is possible to provide nonsense arguments to hamctl commands
leading to unexpected behaviour. As an example it is possible to run the
following 'release' command even though a 'describe' argument is passed.

    $ hamctl release describe --branch master --env prod
    Release of service fee-statement using branch master
    [✓] Release of master-49f970f6fa-2672319eac to prod initialized

This change adds a guard on all commands that does not use arguents to not allow
any.

    $ hamctl release describe --branch master --env prod
    Error: accepts 0 arg(s), received 1


kaspernissen (approved):
Nice 🚀

PR #238, opened 2021-04-15
Only send events when k8s resource has all needed labels

jweibel22:


cmd/daemon/kubernetes/kubernetes.go
```diff
@@ -34,10 +34,11 @@ func NewClient(kubeConfigPath string, moduloCrashReportNotif float64, e Exporter
 }
 
 func isCorrectlyAnnotated(annotations map[string]string) bool {
-	if !(annotations["lunarway.com/controlled-by-release-manager"] == "true") &&
-		annotations["lunarway.com/artifact-id"] == "" &&
-		annotations["lunarway.com/author"] == "" {
-		return false
+
```
mahlunar, line 37: Remove new line?

mahlunar: Also applicable for kubernetes_test.go

mahlunar (approved):

cmd/daemon/kubernetes/kubernetes_test.go
```diff
@@ -0,0 +1,26 @@
+package kubernetes
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestMisingLabelsIsNotCorrectlyAnnotated(t *testing.T) {
```
Crevil, line 8: This is a case where table tests are the common choice in Go.
To me at least, this makes the cases tested more clear.

```go
func TestIsCorrectlyAnnotated(t *testing.T) {
	tt := []struct {
		name       string
		controlled string
		artifactID string
		author     string
		correct    bool
	}{
		{
			name:       "only controlled",
			controlled: "true",
			artifactID: "",
			author:     "",
			correct:    false,
		},
		{
			name:       "all good",
			controlled: "true",
			artifactID: "1",
			author:     "platon",
			correct:    true,
		},
	}
	for _, tc := range tt {
		t.Run(tc.name, func(t *testing.T) {
			correct := isCorrectlyAnnotated(map[string]string{
				"lunarway.com/controlled-by-release-manager": tc.controlled,
				"lunarway.com/artifactId":                    tc.artifactID,
				"lunarway.com/author":                        tc.author,
			})

			assert.Equal(t, tc.correct, correct)
		})
	}
}
```

Crevil (approved):

cmd/daemon/kubernetes/kubernetes.go
jweibel22: I usually add a newline before EOF because it plays well with git diffs. my go-land go formatter doesn't complain about them. is this not "go idiomatic"?

Crevil: Its not the EOF newmine but the function opening I think. :)

jweibel22: ahh :-)

mahlunar: I remember pointing the function opening. weird :)

Crevil: The love of rebasing during review @jweibel22 ? 🙈

jweibel22:
@Crevil there seem to be some flakiness with the integration-tests :-( but I guess you're aware?

Crevil:
Yeahh... It sucks and we know. :(

PR #241, opened 2021-05-10
Fix possible insecure path joins

hoeg:
Currently function arguments are passed directly to `path.Join(...)` which might lead to directory traversal attacks due to these arguments not being sanitized for e.g. `../` (see https://play.golang.org/p/qqkltPKjYj3)

### DIFF
--- a/internal/policy/policy.go
+++ b/internal/policy/policy.go
@@ -225,7 +225,7 @@ func (s *Service) updatePolicies(ctx context.Context, actor Actor, svc, commitMs
 
 		policiesPath, err := securejoin.SecureJoin(policiesDir, fmt.Sprintf("%s.json", svc))
 		if err != nil {
-			return true, errors.WithMessage(err, "join policy path")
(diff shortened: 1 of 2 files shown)


internal/policy/policy.go
```diff
@@ -223,7 +223,10 @@ func (s *Service) updatePolicies(ctx context.Context, actor Actor, svc, commitMs
 			return true, errors.WithMessagef(err, "make policies directory '%s'", policiesDir)
 		}
 
-		policiesPath := path.Join(policiesDir, fmt.Sprintf("%s.json", svc))
+		policiesPath, err := securejoin.SecureJoin(policiesDir, fmt.Sprintf("%s.json", svc))
+		if err != nil {
+			return true, errors.WithMessage(err, "join policy path")
```
Crevil, line 228: Should we add the `policiesDir` to this error to understand what happened?

Crevil (approved):

internal/policy/policy.go
hoeg: Yes, more context is a good idea 👍

PR #242, opened 2021-06-21
Add branch filter to describe artifacts

Crevil:
This change adds a branch filter to `/describe/artifacts` that can be used to only
get artifacts from a specific branch.

To test the otherwise untested `ArtifactSpecifications` method a new dependency is
added for faking S3 storage. This makes it trivial to seed a bucket with test
data and use that in tests. The added test `TestService_ArtifactSpecifications`
covers existing functionality along with the new branch filter.

To make the `artifactreadstoreage.Service` testable the AWS clients and session is
extracted out of the constructor and instead injected in start.go on the server.


cmd/server/command/start.go
```diff
@@ -169,7 +173,18 @@ func NewStart(startOptions *startOptions) *cobra.Command {
 
 			var s3storageSvc *s3storage.Service
 			if startOptions.s3storage.S3BucketName != "" {
-				s3storageSvc, err = s3storage.New(startOptions.s3storage.S3BucketName, tracer)
+				region := "eu-west-1"
```
mahlunar, line 176: I think we should make this one configurable?

mahlunar (approved):

cmd/server/command/start.go
Crevil: I didn't want to change anything in this behaviour, but I agree. Will follow up with some additional fixes I think.

Another things is to replace the existing integration tests with the new fake S3 client to make them runable in CI.

Crevil: I prepared the removal of the live S3 fixtures and it turns out it was super simple so I have added that to this branch.

PR #243, opened 2021-06-22
Remove flux event feature

Crevil:
Currently the release-daemon can listen for flux events over HTTP and
websockets. We have disabled this functionality in all environments and are in
the process of migrating to flux v2 as well so the feature is removed.


kaspernissen (approved):
Great 👍

PR #245, opened 2021-06-22
Fix unmarshaled artifact stage types

Crevil:
Currently when reading an artifact spec from JSON all stage's Data field becomes
a generic map[string]interface{} as they are typed with interface{}. This makes
them hard to work with as we do in the status flow where we sum up
vulnerabilities.

This was discovered in a more challenging way when introducing a new HTTP
transport layer where we have static transport DTOs instead of reusing the
internal artifact.Spec on the transport level. That change will be provided in a
future PR.

This change adds a custom unmarshaling method for the Stage type which detects
the stage variant by its ID and then unmarshals the data into a concrete type.


kaspernissen (approved):
\lgtm

PR #246, opened 2021-06-22
Align hamctl template usage and decople data model

Crevil:
Currently we have three ways of writing output to the screen in hamctl. `1)`
`fmt.Print*()` variants for simple line outputs, `2)` `fmt.Print*()` variants for complex
outputs and `3)` Go templates.

`1)` is ok for the most simple outputs and are left as is. `2)` is hard to read and
modify and are updated to use templates instead.

All the templates are updated to use a dedicated data model to avoid coupling
HTTP DTOs to the outputs which is error prone and hard to catch at compile time.

### DIFF
--- a/cmd/hamctl/command/describe.go
+++ b/cmd/hamctl/command/describe.go
@@ -2,12 +2,15 @@ package command
 
 import (
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
 	"os"
+	"time"
 
 	"github.com/lunarway/release-manager/cmd/hamctl/command/actions"
 	"github.com/lunarway/release-manager/cmd/hamctl/command/completion"
+	"github.com/lunarway/release-manager/cmd/hamctl/template"
 	httpinternal "github.com/lunarway/release-manager/internal/http"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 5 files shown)


cmd/hamctl/command/describe.go
```diff
@@ -74,7 +101,7 @@ Format the output with a custom template:
 			if len(template) == 0 {
 				template = describeReleaseDefaultTemplate
 			}
-			err = templateOutput(os.Stdout, "describeRelease", template, releasesResponse)
+			err = templateDescribeRelease(os.Stdout, template, mapReleaseResponseToTemplate(releasesResponse))
```
jweibel22, line 104: A curious question (unrelated to the changes in this PR, but this is the first time I see the code) I can see how a making it possible to use a custom template makes it very flexible, but it also couples the clients to the DTO defined here. It might be very stable so maybe not a problem but I think we should add a comment that it is in fact an interface, and thus shouldn't be changed in a breaking way. Also, how does a client know what that interface looks like (i.e. what he can use in his custom template), he would have to go look at the code here? It should probably be included in the help description.

cmd/hamctl/command/describe.go
```diff
@@ -32,20 +35,44 @@ Get details about the current release of product in the dev environment:
...
 {{ end }}
 `
 
+type describeReleaseData struct {
+	Service     string
+	Environment string
+	Releases    []describeReleaseDataRelease
+}
+
+type describeReleaseDataRelease struct {
```
jweibel22, line 58: Kudos for defining dedicated DTOs here 🚀

cmd/hamctl/command/describe.go
Crevil: Yes, that is a "classic" with templated commands, no one ever knows what the datamodel for it is.

A better help description would be nice to have and this change makes that very essy and way more stable. I'll see if I can follow up with that.

jweibel22 (approved):
Creating dedicated DTOs for the CLIs and aligning the printing are major improvements!
It will make it much easier to maintain this code going forward 🚀 

Just checking: Since there are no tests, the development process here is to run the CLI locally from the terminal while making changes (and point it to a local server or the deployed release manager)? I think that is ok for now if it's easy to start it up locally.

Crevil:
Yes, that is it for now. It got late last night, but it isn't hard to test these templates kn unit tests. We can start the server with `make server-memory` and point to it, or point to a live server somewhere.

I'll follow up with some tests on these 🙌

PR #247, opened 2021-06-23
Add tests of hamctl status template

Crevil:
Add unit tests of the status template to ensure it renderes correctly. It did
not, so there are several small changes to both the look and layout.

Along the way a bug was found in the commonly used format string so this is
extracted out to a shared template variable to control how dates are presented.

Follow up on #246


mahlunar (approved):
Nice 🎉

Crevil:
FYI @jweibel22

jweibel22:
> FYI @jweibel22

Awesome 🚀

PR #257, opened 2021-08-28
Add dynamic Grafana annotation flag

Crevil:
Currently Grafana annotations are hard coded with dedicated flags for each
supported environment. This makes the code hard coupled and cumbersome to expand
with new environments.

This change adds a new `grafana-annnotations` flag that accepts a comma
separated list of `<env>=<api-key>=<url>` values. This makes it possible to add
as many environments as needed without making changes to the code base.

The `grafanaOptions` struct impements the `pflag.Value` and `pflag.SliceValue`
interfaces which makes `pflag` handle multiple apperances of the flag correctly.
This means the following flag settings are all valid and will result in 4
configured environments.

```
server start \
  --grafana-annotations dev=${GRAFANA_API_KEY}=${GRAFANA_URL} \
  --grafana-annotations staging=${GRAFANA_API_KEY}=${GRAFANA_URL} \
  --grafana-annotations prod=${GRAFANA_API_KEY}=${GRAFANA_URL},platform=${GRAFANA_API_KEY}=${GRAFANA_URL}
```

### DIFF
--- a/cmd/server/command/flags.go
+++ b/cmd/server/command/flags.go
@@ -4,8 +4,20 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/lunarway/release-manager/internal/grafana"
 	"github.com/pkg/errors"
 )
+
+func mapGrafanaOptionsToEnvironment(opts *grafanaOptions) map[string]grafana.Environment {
+	environments := make(map[string]grafana.Environment)
+	for env, config := range *opts {
+		environments[env] = grafana.Environment{
+			APIKey:  config.APIKey,
+			BaseURL: config.URL,
+		}
+	}
+	return environments
+}
 
 type grafanaConfig struct {
 	URL    string
--- a/cmd/server/command/flags.go
+++ b/cmd/server/command/flags.go
@@ -1,0 +1,82 @@
+package command
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+type grafanaConfig struct {
+	URL    string
+	APIKey string
+}
+
+type grafanaOptions map[string]grafanaConfig
+
+func (opts *grafanaOptions) String() string {
+	values := opts.GetSlice()
+	if len(values) == 0 {
+		return "[]"
+	}
+	return strings.Join(values, ",")
+}
+
+func (opts *grafanaOptions) Type() string {
+	return "<env>=<api-key>=<url>"
+}
+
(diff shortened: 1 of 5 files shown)


cmd/server/command/flags.go
```diff
@@ -0,0 +1,82 @@
+package command
```
mahlunar, line 1: I think the file name should be changed or will it handle other stuff than Grafana?

cmd/server/command/start.go
```diff
@@ -456,3 +434,14 @@ func getBroker(c *brokerOptions) (broker.Broker, error) {
 		return nil, errors.New("no broker selected")
 	}
 }
+
+func mapGrafanaOptionsToEnvironment(opts *grafanaOptions) map[string]grafana.Environment {
```
mahlunar, line 438: Consider moving this function into the file with the other grafanaOptions

cmd/server/command/flags.go
```diff
@@ -0,0 +1,82 @@
...
+func (opts *grafanaOptions) GetSlice() []string {
+	var values []string
+	for env, value := range *opts {
+		values = append(values, fmt.Sprintf("%s=<redacted>=%s", env, value.URL))
+	}
+	return values
+}
+
+// Append adds the specified value to the end of the flag value list.
+func (opts *grafanaOptions) Append(value string) error {
```
mahlunar, line 49: This function seems to missing tests?

cmd/server/command/flags.go
```diff
@@ -0,0 +1,82 @@
...
+	env, config, err := parseGrafanaConfig(value)
+	if err != nil {
+		return err
+	}
+	(*opts)[env] = config
+	return nil
+}
+
+// Replace will fully overwrite any data currently in the flag value list.
+func (opts *grafanaOptions) Replace(values []string) error {
```
mahlunar, line 59: This function seems to be missing tests?

cmd/server/command/flags.go
Crevil: I'm planning on moving the other custom flags over here as well :)

cmd/server/command/start.go
Crevil: 👍

cmd/server/command/flags.go
Crevil: True. I didn't do that as it seems super simple 😉  I'll add them 🎉

mahlunar (approved):

PR #260, opened 2021-11-05
Ignore synthetic ADDED events in daemon

Crevil:
Currently the daemon will process the [synthetic `ADDED` events](https://kubernetes.io/docs/reference/using-api/api-concepts/#semantics-for-watch) that are emitted
from a watch intended to create the intial state. This results in the daemon
notifying on every pod, deployment job etc. that are in a cluster when it starts
up the first time. If slack messages are enabled in the release-manager this
results in a whole lot of messages hitting developers.

This is especially painful when releasing the daemon into a new cluster that
hasn't been observed before. For Lunar this is often failover clusters where the
daemon is disabled until a failover cluster has settled its state.

This change disables these events by replacing the low level watch primitive on
`kubernetes.Clientset` to use the higher level abstraction `informers`.
Informers can tell when they have synced the inital state and with this we have
a hook as to when we should start forwarding events to the release-manager. This
also has the benifit of providing a simpler implementation without excessive use
of channels and Go routines as informers just register callbacks on specific
resources.

The diff is quite large but no functionality is changed in the handlers but they
are instead moved to be separate structs instead of methods on the
`kubernetes.Client` struct.

### DIFF
--- a/cmd/daemon/command/start.go
+++ b/cmd/daemon/command/start.go
@@ -39,7 +39,7 @@ func StartDaemon() *cobra.Command {
 				Environment: environment,
 			}
 
-			kubectl, err := kubernetes.NewClient(kubeConfigPath, moduloCrashReportNotif, exporter)
+			kubectl, err := kubernetes.NewClient(kubeConfigPath)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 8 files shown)


kaspernissen (approved):
\lgtm

PR #263, opened 2021-11-10
Annotate Jobs when observing them

Crevil:
Currently if a job gets into a Failed state and the is a fail history we will
keep sending JobError events to the server over and over.

This change adds the observed annotation to Jobs when we handle them to only
ever send a single even for a failed Job.

This is the same as done doe DaemonSets and Deployments so the code is
refactored to be more shareable and to avoid duplicating the annotation keys all
over.

### DIFF
--- a/cmd/daemon/kubernetes/jobs.go
+++ b/cmd/daemon/kubernetes/jobs.go
@@ -7,17 +7,21 @@ import (
 	"github.com/lunarway/release-manager/internal/log"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 )
 
 type JobInformer struct {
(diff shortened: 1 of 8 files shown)


cmd/daemon/kubernetes/jobs.go
```diff
@@ -54,14 +62,30 @@ func (j JobInformer) handle(e interface{}) {
 			JobName:     job.Name,
 			Namespace:   job.Namespace,
 			Errors:      jobErrorMessages(job),
-			ArtifactID:  job.Annotations["lunarway.com/artifact-id"],
+			ArtifactID:  job.Annotations[artifactIDAnnotationKey],
 			AuthorEmail: job.Annotations["lunarway.com/author"],
```
mahlunar, line 66: I believe it would be a good idea to extract this value to a const as well. It is set in multiple places and as the`artifactIDAnnotationKey` is also a const then we should align them.

WDYT?

mahlunar (approved):
🚀 

Just a small comment

cmd/daemon/kubernetes/jobs.go
Crevil: Yeah lets follow up with that. I only changed the other one as I duplicated its use with this change 🙌

PR #272, opened 2021-12-15
Add template functions to rollback templates

Crevil:
Currently the hamctl rollback command relies on a `dateFormat` function that is
not available. This is a bug introduced in
9354fb3854be6fa1c85e756f82cff6e04f39f06c (#247) where the `dateFormat` function
was introduced. It was not however ensured that it was available in the
interactive templates.

This change fixes the issue by creating a new `FuncMap` function that returns all
the functions used in templates and add them to both the `template.Output`
function and the interactive templates.

### DIFF
--- a/cmd/hamctl/command/rollback.go
+++ b/cmd/hamctl/command/rollback.go
@@ -6,7 +6,6 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


mahlunar (approved):
🚀

PR #276, opened 2022-01-05
Revert "Use Christmas emoji (#271)"

mahlunar:
It is not Christmas anymore so let's revert back to Sonic

This reverts commit 659560151b6ca24eaaea2f65f492b3945feea569.


Crevil (approved):
I love how we just hard code this stuff 🙈

mahlunar:
> I love how we just hard code this stuff 🙈 

Yes, we should avoid that going forward.

PR #278, opened 2022-01-20
Ignored a case where configmap cache sync times out

mahlunar:
Currently, we sometimes get an error like the following:

image

This change adds a filter to ignore that error as it is transient failure


kaspernissen (approved):
Nice 🚀

PR #279, opened 2022-03-02
Remove staging in promote flows

Crevil:
As we have discontinued the concept of a staging environment the hardcoded
promotion flow of master -> dev -> staging -> prod no longer makes sense. A long
term change on how promotes are done is in the pipeline but until that is done
this change short circuits the promotion flow to skip staging.

This change also removes any mention of staging in the code base to avoid
confusion. Any code currently presenting something around the staging name is
also updated like the hamctl status command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 GitOps release manager for kubernetes configuration repositories.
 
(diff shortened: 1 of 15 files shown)


mahlunar (approved):
🚀

PR #280, opened 2022-03-02
Enable dependabot updates

Crevil:
This change enables dependabot to keep Go modules, Dockerfiles and GitHub action
workflows up to date.

PR #281, opened 2022-03-02
Migrate to GitHub actions

Crevil:
This change migrates the CI and CD pipeline to GitHub actions.
The build workflow does the same as Travis CI setup.


.github/workflows/build.yml
```diff
@@ -0,0 +1,53 @@
+name: build
+
+on:
+  push:
+    branches:
+      - "master"
+    tags:
+      - "v*"
+  pull_request:
```
mahlunar, line 9: Is this right? Should there not be a value?

mahlunar (approved):
This is awesome. Let's also add Dependabot support for go.mod and github actions

.github/workflows/build.yml
Crevil: It is from their docs so I guess it's good: https://goreleaser.com/ci/actions/

Crevil:
You gonna love #280 @mahlunar 🚀

Crevil:
I realize that this change actually adds a new feature as well. Caching of Go modules 🚀

mahlunar:
> I realize that this change actually adds a new feature as well. Caching of Go modules 🚀

Great!

.github/workflows/build.yml
```diff
@@ -0,0 +1,53 @@
...
+      - name: Tests
+        run: |
+          make integration-test VERBOSE=yes
+      - name: Docker Login
+        uses: docker/login-action@v1
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_ROBOT_TOKEN }}
+      - name: Run GoReleaser
```
mahlunar, line 45: We might be able to run go-release in a separate action if we want? To avoid cluttering this one

Crevil: Yeah it might make sense. I'm in unknown territory here so let's refactor as makes sense 👍

mahlunar: We learn as we go!

Crevil:
Great I got a green build 🚀  #282 contains the above commits. Once that is merged I will rebase this branch so we can get actions in with a clean commit.

PR #282, opened 2022-03-02
Remove dependency on host git in tests

Crevil:
Currently some of the tests rely on valid git configuration being pressent when
running. With the migration to GitHub actions in #281 this became clear as the
config was not pressent.

This change adds two new non-documented environment variables that can be used
to "fake" the user name and email for use in tests. This is not a great solution
but in generel the git config lookup isn't a great solution so I figure this is
OK for now. Another test case that makes some initial commits have also been
updated to set the commit author to again remove the dependency on the host
config.

The build in #281 is green with these changes is so we should be good in that
nothing else depends on git in this way.


mahlunar (approved):
🚀  LGTM

PR #365, opened 2022-05-27
Webhook to release-daemon for flux2-notification alerts

MariaCFFrandsen:
This PR adds a webhook by exposing a http endpoint for flux2-notification-controller to post alerts through. This is step towards giving developers more insight into errors in different env's. This includes a unit-test of endpoint with a request body that conforms to documentation.


cmd/daemon/command/start.go
```diff
@@ -93,5 +94,7 @@ func StartDaemon() *cobra.Command {
 	//nolint:errcheck
 	command.MarkFlagRequired("environment")
 	logConfiguration = log.RegisterFlags(command)
+
+	flux_notification_controller.StartHttpServer()
```
Crevil, line 98: This needs to go into the `RunE` function above along with the rest of the "kubernetes watcher" stuff. That way it is started in the lifetime of the general application 😄  So just before the `Deamon started` log line maybe.

cmd/daemon/flux-notification-controller/api.go
```diff
@@ -0,0 +1,35 @@
...
+	ReportingController string    `json:"reportingController"`
+	ReportingInstance   string    `json:"reportingInstance"`
+}
+
+func HandleEventFromFlux2(w http.ResponseWriter, r *http.Request) {
+	defer r.Body.Close()
+	body, _ := ioutil.ReadAll(r.Body) //TODO: log something
+	var event Event
+	_ = json.Unmarshal(body, &event)
+	zlog.Infof("Received alert from flux2-notification-controller: %s with msg: %s", event.InvolvedObject.Name, event.Message)
```
Crevil, line 34: Notice that `zlog` is not open source so we cannot use it here. Instead use the internal `github.com/lunarway/release-manager/internal/log` package.

```suggestion
	log.Infof("Received alert from flux2-notification-controller: %s with msg: %s", event.InvolvedObject.Name, event.Message)
```

cmd/daemon/flux-notification-controller/server.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"net/http"
+
+	"github.com/gorilla/mux"
+)
+
+func StartHttpServer() {
+	router := mux.NewRouter()
+	router.HandleFunc("/webhook/flux2-alerts", HandleEventFromFlux2).Methods(http.MethodPost)
+	_ = http.ListenAndServe(":3001", router) //TODO: log something
```
Crevil, line 12: Let's get an error log in here from the get go 😄 

```suggestion
	err := http.ListenAndServe(":3001", router) //TODO: log something
	if err != nil {
		log.Errorf("Failed to start HTTP server: %v", err)
	}
```

Crevil (changes requested):

cmd/daemon/flux-notification-controller/server.go
```diff
@@ -0,0 +1,15 @@
...
+
+	"github.com/gorilla/mux"
+	"github.com/lunarway/release-manager/internal/log"
+)
+
+func StartHttpServer() {
+	router := mux.NewRouter()
+	router.HandleFunc("/webhook/flux2-alerts", HandleEventFromFlux2).Methods(http.MethodPost)
+	err := http.ListenAndServe(":3001", router)
+	log.Errorf("Failed to start daemon's HTTP server: %v", err)
```
Crevil, line 14: To avoid logging `Failed to start daemon's HTTP server: nil` when the server stops we need a condition around this.

```suggestion
	if err != nil {
		log.Errorf("Failed to start daemon's HTTP server: %v", err)
	}
```

cmd/daemon/flux-notification-controller/api.go
```diff
@@ -0,0 +1,35 @@
...
+	ReportingController string    `json:"reportingController"`
+	ReportingInstance   string    `json:"reportingInstance"`
+}
+
+func HandleEventFromFlux2(w http.ResponseWriter, r *http.Request) {
+	defer r.Body.Close()
+	body, _ := ioutil.ReadAll(r.Body) //TODO: log something
+	var event Event
+	_ = json.Unmarshal(body, &event)
+	log.Infof("Received alert from flux2-notification-controller: %s with msg: %s", event.InvolvedObject.Name, event.Message)
```
Crevil, line 34: Should we get the logging in place here up front to avoid more turn around if something works unexpectedly?
The most common case of errors here would be that our `Event` types doesn't match what we receive.

Consider this change. I add logging and an error response to let the caller know something went wrong and we get some context to it as well. WDYT?

```suggestion
	body, err := ioutil.ReadAll(r.Body) //TODO: log something
	if err != nil {
		log.Errorf("Failed to unmarshal alert from flux2-notification-controller: %v", err)
		http.Error(w, "unknown error", http. StatusInternalServerError)
		return
	}
	
	var event Event
	err = json.Unmarshal(body, &event)
	if err != nil {
		log.Errorf("Failed to unmarshal alert from flux2-notification-controller: %v", err)
		http.Error(w, "unknown error", http. StatusInternalServerError)
		return
	}
	log.Infof("Received alert from flux2-notification-controller: %s with msg: %s", event.InvolvedObject.Name, event.Message)
```

We _could_ consider adding a payload log to ease debugging as well in the unmarshal case to have the "real" payload logged out. Something like `log.Infof("Alert payload: %s", body)`

cmd/daemon/flux-notification-controller/server.go
```diff
@@ -0,0 +1,15 @@
+package flux_notification_controller
```
Crevil, line 1: A nit, but usually we don't have multi word package names. I'm at a loss for coming up with a better name right now through, so let's keep it as is 🙈

cmd/daemon/flux-notification-controller/server.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"net/http"
+
+	"github.com/gorilla/mux"
```
Crevil, line 6: Any reason for introducing `mux` here as a dependency compared to the standard libraries `http.ServeMux`?

Crevil (changes requested):

cmd/daemon/flux-notification-controller/server.go
MariaCFFrandsen: 🙈

cmd/daemon/flux-notification-controller/server.go
MariaCFFrandsen: flux2noticatons would be an alternative? :)

Crevil: Yeah at least that avoids the `_` 👍

cmd/daemon/flux2notifications/api_test.go
```diff
@@ -0,0 +1,42 @@
...
+	)
+
+	//act
+	HandleEventFromFlux2(w, request)
+	response := w.Result()
+	defer response.Body.Close()
+	_, err := ioutil.ReadAll(response.Body)
+
+	//assert
+	assert.NoError(t, err, "HandleEventFromFlux2 could not handle request")
```
Crevil, line 32: Should we assert on the status code as well here? Otherwise we only check if we can read the body?

Crevil (approved):
🚀

PR #369, opened 2022-06-09
Notify on pod OOMKilled statuses

mahlunar:
Currently, developers does not get information about `OOMKilled` status updates.

This change introduces the check and sends a pod error event.


Crevil (approved):
🚀🚀🚀🚀🚀🚀

PR #380, opened 2022-06-24
Feature/web hook amqp

mahlunar:
Currently, when running in multiple replicas, pods can have stale data due to a Github web hook only being received in one of the instances. 

This change adds AMQP  fanout so that Github web hooks are forwarded to each replica to mitigate staleness.


mahlunar:
I am unsure how to do a proper end to end test. I think that is needed here to avoid breaking flows

go.mod
```diff
@@ -32,7 +32,10 @@ require (
 	k8s.io/client-go v0.24.3
 )
 
-require github.com/gorilla/mux v1.8.0
+require (
+	github.com/gorilla/mux v1.8.0
+	k8s.io/utils v0.0.0-20220210201930-3a6ce19ff2f9
+)
```
Crevil, line 38: Is this change required or something from testing?

internal/broker/amqpextra/amqpextra.go
```diff
@@ -46,6 +48,9 @@ func New(c Config) (*Worker, error) {
 		return nil, err
 	}
 
+	Exchange = c.Exchange
+	ExchangeFanout = fmt.Sprintf("%s-git-fanout", c.Exchange)
```
Crevil, line 52: Should this be `-github-fanout` just for explicitness?

internal/broker/amqpextra/consumer.go
```diff
@@ -15,27 +18,48 @@ import (
...
+			Exchange:            ExchangeFanout,
+			DurableExchange:     false,
+			AutoDeletedExchange: true,
+			Queue:               fmt.Sprintf("%s-git-fanout-%s", w.config.Queue, fanoutID),
+			ExclusiveQueue:      true,
+			DurableQueue:        false,
+			Prefetch:            0,
+			RoutingPatterns:     []string{"#"},
+			Handle: func(message *amqp.Delivery) error {
+				return nil
```
Crevil, line 55: Shouldn't we call the fanout handlers here somewhere?

Crevil (commented):
Overall I think this is great. As discussed I think we should see if we can get a test in here to verify the functionality.

go.mod
mahlunar: `go mod tidy` did something strange here. Both dependencies were there before. I have merged them back in to the original list.

mahlunar:
@Crevil I figured out how to test this locally and can see I have a connection from the fanout exchange to the local queue.

Please review again and let's discuss how we could automate tests as well.

Crevil (approved):

PR #382, opened 2022-07-04
Removes ticks when error message is empty

MariaCFFrandsen:
An OOMKIlled error seems to have an empty error msg. Therefore, it looks a bit weird in Slack:
image

When @mahlunar made the task, he also needed a reference to the pods.go -> [release-manager/pods.go at master · lunarway/release-manager](https://github.com/lunarway/release-manager/blob/master/cmd/daemon/kubernetes/pods.go#L143). I can't really figure out why this is relevant for the Slack message.

I am also unsure of how to test it. Any guidance would be helpful 🙏 It seems like the rest of the methods were tested through mock but this method is not part of the interface and it would be a mistake to it purely for testing purposes.

### DIFF
--- a/internal/slack/slack.go
+++ b/internal/slack/slack.go
@@ -208,7 +208,7 @@ func (c *Client) NotifySlackPolicyFailed(ctx context.Context, email, title, erro
 	attachments := slack.MsgOptionAttachments(slack.Attachment{
 		Title:      title,
 		Color:      MsgColorRed,
-		Text:       FormatErrorMessage(errorMessage),
+		Text:       formatErrorMessage(errorMessage),
 		MarkdownIn: []string{"text", "fields"},
 	})
 	_, _, err = c.client.PostMessageContext(ctx, userID, asUser, attachments)
@@ -298,7 +298,7 @@ func (c *Client) NotifyK8SPodErrorEvent(ctx context.Context, event *http.PodErro
 	for _, container := range event.Errors {
 		fields = append(fields, slack.AttachmentField{
 			Title: fmt.Sprintf("Container: %s (%s)", container.Name, container.Type),
-			Value: fmt.Sprintf("```%s```", container.ErrorMessage),
+			Value: formatErrorMessage(container.ErrorMessage),
 			Short: false,
 		})
 	}
@@ -329,7 +329,7 @@ func (c *Client) NotifyK8SJobErrorEvent(ctx context.Context, event *http.JobErro
 	for _, condition := range event.Errors {
 		fields = append(fields, slack.AttachmentField{
 			Title: fmt.Sprintf("Reason: %s", condition.Reason),
-			Value: fmt.Sprintf("```%s```", condition.Message),
+			Value: formatErrorMessage(condition.Message),
 			Short: false,
 		})
 	}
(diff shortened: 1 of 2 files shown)


internal/slack/slack.go
```diff
@@ -208,10 +208,11 @@ func (c *Client) NotifySlackPolicyFailed(ctx context.Context, email, title, erro
 	attachments := slack.MsgOptionAttachments(slack.Attachment{
 		Title:      title,
 		Color:      MsgColorRed,
-		Text:       fmt.Sprintf("```%s```", errorMessage),
+		Text:       FormatErrorMessage(errorMessage),
 		MarkdownIn: []string{"text", "fields"},
 	})
-
+	log.Infof("errormsg = %s", FormatErrorMessage(errorMessage))
+	fmt.Println(FormatErrorMessage(errorMessage))
```
Crevil, line 215: I guess this is from debugging?

internal/slack/slack.go
```diff
@@ -208,10 +208,11 @@ func (c *Client) NotifySlackPolicyFailed(ctx context.Context, email, title, erro
 	attachments := slack.MsgOptionAttachments(slack.Attachment{
 		Title:      title,
 		Color:      MsgColorRed,
-		Text:       fmt.Sprintf("```%s```", errorMessage),
+		Text:       FormatErrorMessage(errorMessage),
 		MarkdownIn: []string{"text", "fields"},
 	})
-
+	log.Infof("errormsg = %s", FormatErrorMessage(errorMessage))
```
Crevil, line 214: Do we need to log this?

internal/slack/slack_test.go
```diff
@@ -36,21 +36,23 @@ func TestNewMutableClient_mapping(t *testing.T) {
...
-	slackClient.Test(t)
-	slackClient.On("PostMessage", slackUser.ID, mock.Anything, mock.Anything).Return("", "", nil)
-	// return error when looking up non-corp email
-	slackClient.On("GetUserByEmailContext", mock.Anything, email).Return(&slack.User{}, errors.New("users_not_found"))
-	// return slack user on anything else
-	slackClient.On("GetUserByEmailContext", mock.Anything, mock.Anything).Return(slackUser, nil)
+	t.Run("domain verification", func(t *testing.T) {
+		slackClient := intslack.MockSlackClient{}
+		slackClient.Test(t)
+		slackClient.On("PostMessage", slackUser.ID, mock.Anything, mock.Anything).Return("", "", nil)
```
Crevil, line 42: You can make an assertion on the payload of the posted message here I think.

Crevil (changes requested):

internal/slack/slack_test.go
MariaCFFrandsen: I am not sure how to do that :) I am interested in `slack.MsgOptionAttachments` but these are mocked so no actually logic concerning them is not performed.

Crevil: Ah I see. The `slack.MsgOption` option pattern makes it hard to assert on what is provided. 👍 No worries.

Crevil:
I just reread this PR and there are a couple of more places we need to handle this.

- https://github.com/lunarway/release-manager/blob/master/internal/slack/slack.go#L302
- https://github.com/lunarway/release-manager/blob/master/internal/slack/slack.go#L333

Crevil (changes requested):
See above comment 🙈

Crevil (approved):
🚀

PR #388, opened 2022-07-15
fix getCodeOwnerSquad

MariaCFFrandsen:
[garbage was logged ](https://humio.lunar.tech/deployment/search?query=%22CrashLoopBackOff%20owned%20by%22%20%0A&live=false&start=24h) due the function was printed out instead of the result. 
log.Infof("Pod: %s is in CrashLoopBackOff owned by squad %s", pod.Name, getCodeOwnerSquad)

right:
log.Infof("Pod: %s is in CrashLoopBackOff owned by squad %s", pod.Name, getCodeOwnerSquad(pod.Labels))
Took a while to debug 🙈

### DIFF
--- a/cmd/daemon/kubernetes/pods.go
+++ b/cmd/daemon/kubernetes/pods.go
@@ -70,7 +70,7 @@ func (p *PodInformer) handle(e interface{}) {
 				pod.Name)
 			return
 		}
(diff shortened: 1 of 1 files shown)


JacobValdemar (approved):
Nice find :)

PR #389, opened 2022-07-15
send crashloopbackoff events into platform squads alert channel

MariaCFFrandsen:


JacobValdemar (approved):
*_*

internal/slack/slack.go
```diff
@@ -313,6 +313,9 @@ func (c *Client) NotifyK8SPodErrorEvent(ctx context.Context, event *http.PodErro
 	if err != nil {
 		return err
 	}
+	if event.Squad == "aura" || event.Squad == "odyssey" {
```
EhsanLunar, line 316: This is a public repo, should we really have our very specific squad names hardcoded into it?

MariaCFFrandsen: I would like to test out in our own alerts channel in case it doesn't work as intended so I don't prematurely spam everybody.

EhsanLunar: Would be nice, if it works, to allow this to check if the squad has opted-in for alerts and than send it

MariaCFFrandsen: Yes 👍

PR #392, opened 2022-07-19
Refactor rollback to be testable

MariaCFFrandsen:
Currently, the hamctl numbers the rollback starting from zero. This result is some strange behaviour as by @kjuulh, therefore they should be numbered starting from 1.


cmd/hamctl/command/rollback.go
```diff
@@ -48,7 +48,7 @@ has no effect.`,
 					return err
 				}
 
-				if len(releasesResponse.Releases) < 2 {
+				if len(releasesResponse.Releases) < 2 { //should there just be 1 release to do rollback? so len(releasesResponse.Releases) < 1?
```
MariaCFFrandsen, line 51: WDUT?

JacobValdemar: What do you roll back to if you are rolling back the only release? 🤔

MariaCFFrandsen: Maybe I assume wrong but is the current release all part of the rollback options?

cmd/hamctl/command/rollback.go
```diff
@@ -62,7 +62,7 @@ has no effect.`,
 
 				searcher := func(input string, index int) bool {
 					release := items[index]
-					name := strings.ToLower(fmt.Sprintf("#%v %s", release.ReleaseIndex, release.ArtifactID))
+					name := strings.ToLower(fmt.Sprintf("#%v %s", release.ReleaseIndex+1, release.ArtifactID))
```
Crevil, line 65: I think this needs explanation. Why the +1?

MariaCFFrandsen: It's to fix the UI: 
image

Crevil: I think we should add a comment in the code to explain this, eg. Using index 1 to make a more semantic listing to users.

MariaCFFrandsen: Would we also need to change the[ input index](https://github.com/lunarway/release-manager/blob/master/cmd/hamctl/command/rollback.go#L80), right?

MariaCFFrandsen:
Convert to a draft

kjuulh:
I will pick this up tomorrow, if time permits.

kjuulh:
Alright, I've added tests. The general assumption about the tests were primarily because the search functionality didn't pick up the numberings correctly. As it turns out it wasn't because it was off by one. But because it found some numbers in the hashes for the rollbacks. 

I've since reverted the ReleaseIndex+1, so that we now keep a 0 index, as we should. I've kept the tests, as they are somewhat useful, though a bit messy because of how the code is structured. I'd redo it at some point and create a full acceptance suite, but that is for another time. Lets do that when we add the HA functionality.

kjuulh:
I will expand on the strategy pattern in some upcoming prs. I just didn't want this PR to get too out of hand.

cmd/hamctl/command/rollback.go
```diff
@@ -52,36 +56,10 @@ has no effect.`,
...
-					StartInSearchMode: true,
-				}
-
-				index, _, err := prompt.Run()
-
+				index, err := selectReleaseUI(environment, releasesResponse.Releases)
 				if err != nil {
-					fmt.Printf("Rollback cancelled: %v\n", err)
-					os.Exit(1)
+					logger("something went wrong select release a release")
```
Crevil, line 61: This is a change in functionality, right? We would before just write "Rollback cancelled" and any error. If I hit CTRL+C in the prompt I wouldn't expect "Something went wrong"?

cmd/hamctl/command/rollback_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+	gitConfigAPI := git.GitConfigAPIMock{
+		CommitterDetailsFunc: func() (string, string, error) {
+			return "name", "email", nil
+		},
+	}
+
+	releaseClient := actions.NewReleaseHttpClient(&gitConfigAPI, &c)
+
+	runCommand := func(selectRollback command.SelectRollbackRelease, t *testing.T, args ...string) ([]string, error) {
```
Crevil, line 83: Test functions usually have `t` as their first argument to make it super clear they only work inside tests. WDYT?

```suggestion
	runCommand := func(t *testing.T, selectRollback command.SelectRollbackRelease, args ...string) ([]string, error) {
```

cmd/hamctl/command/rollback_test.go
```diff
@@ -0,0 +1,181 @@
...
+		var rollback command.SelectRollbackRelease = func(environment string, releases []internalhttp.DescribeReleaseResponseRelease) (int, error) {
+			return -1, nil
+		}
+
+		output, err := runCommand(rollback, t, "--env", "dev", "--artifact", "artifact-0")
+
+		require.NoError(t, err)
+		assert.Equal(
+			t,
+			[]string{"[✓] Starting rollback of service [some-service-name artifact-0] to %!s(MISSING)\n", "[✓] [released]\n"},
```
Crevil, line 149: Do we really expect `%!s(MISSING)` to be written to the user? 🙈

internal/git/config.go
```diff
@@ -9,16 +9,33 @@ import (
 	"github.com/pkg/errors"
 )
 
+//go:generate moq -out config_mock.go . GitConfigAPI
+
+// GitConfigAPI is an interface to interact with a git config system
+// this makes it possible to extract information from the repository
+// or the local user
+type GitConfigAPI interface {
+	CommitterDetails() (name string, email string, err error)
```
Crevil, line 18: It is somewhat unusualy that the provider of functionality also defines its interface? This would normally go into the call site to make sure it only contains what is relevant to the caller instead of all the things a provider provides.

WDYT?

internal/git/config.go
```diff
@@ -9,16 +9,33 @@ import (
...
+}
+
+type LocalGitConfigAPI struct {
+}
+
+// NewLocalGitConfigAPI provide a GitConfigAPI for a local repository
+func NewLocalGitConfigAPI() GitConfigAPI {
```
Crevil, line 25: Same as above. We normally want to accept interfaces and return concrete types. Is this intentionally doing something else?

internal/git/config.go
```diff
@@ -9,16 +9,33 @@ import (
...
+	return &LocalGitConfigAPI{}
+}
+
 // CommitterDetails returns name and email read for a Git configuration file.
 //
 // Fetching the configuration values are delegated to the git CLI and follows
 // precedence rules defined by Git.
-func CommitterDetails() (string, string, error) {
-	name, err := getValue("user.name", "HAMCTL_USER_NAME")
+func (*LocalGitConfigAPI) CommitterDetails() (name string, email string, err error) {
```
Crevil, line 33: Any reason for using named return types? It always sticks out to me 🙈

Crevil (approved):
Overall looks good. I guess the PR title and description is no longer acurate?

cmd/hamctl/command/rollback.go
kjuulh: Woops. I think it may've been a copy paste error, will fix

internal/git/config.go
kjuulh: I agree. It is usually a good idea to use structs as normal and just wrap it in an interface when needed. I will see what I can do to improve this.

internal/git/config.go
kjuulh: Yeah, I am not really using the name types pr. say, I just use it to document the 'types'. I find it confusing that you can have multiple strings, but they semantically have different domain models. I.e. one is a name, another an email and so on.

cmd/hamctl/command/rollback_test.go
kjuulh: Missed ... on args in my formatter. Should've suspected that it was it...

internal/git/config.go
kjuulh: This is a pattern often used in other languages. However, it does mean that a struct can only implement a single interface. I will move the interface closer to the user, and as such this will change to the concrete type instead.

internal/git/config.go
Crevil: Makes sense. I guess what would be "nice" is to actually introduce a struct to name the fields? I will go a long way to avoid named return values 🙈

kjuulh: It has been done =D

internal/git/config.go
kjuulh: I've moved it closer to the caller, this does introduce some extra boiler plate for the interfaces, as it is now declared where it is used.

internal/git/config.go
kjuulh: Should be moved now, I can see the benefit of this, but maybe it shouldn't be as close to the caller as it is now?

PR #394, opened 2022-07-21
Alert squads on PodErrorEvents

MariaCFFrandsen:
Adds functionality for alerting the squad as well as the author on PodErrorEvents. When the author misses, overlooks or forgets a pod failure, it can be a problem as said pod is failing until noticed or a new version is released.
It is the assumption that this leads to more - and silently - failing pods in our environments than necessary.
This change would make failing pods a collective squad responsibility instead of the individual developer.

### DIFF
--- a/cmd/daemon/kubernetes/kubernetes.go
+++ b/cmd/daemon/kubernetes/kubernetes.go
@@ -81,7 +81,7 @@ const (
 	artifactIDAnnotationKey    = "lunarway.com/artifact-id"
 	authorAnnotationKey        = "lunarway.com/author"
 	controlledAnnotationKey    = "lunarway.com/controlled-by-release-manager"
-	runtimeAlertsAnnotationKey = "lunarway.com/runtimeAlerts"
+	runtimeAlertsAnnotationKey = "lunarway.com/runtime-alerts"
 	squadLabelKey              = "squad"
 )
 
--- a/cmd/daemon/kubernetes/kubernetes.go
+++ b/cmd/daemon/kubernetes/kubernetes.go
@@ -77,11 +77,12 @@ func isCorrectlyAnnotated(annotations map[string]string) bool {
 }
 
 const (
-	observedAnnotationKey   = "lunarway.com/observed-artifact-id"
-	artifactIDAnnotationKey = "lunarway.com/artifact-id"
-	authorAnnotationKey     = "lunarway.com/author"
-	controlledAnnotationKey = "lunarway.com/controlled-by-release-manager"
-	squadLabelKey           = "squad"
+	observedAnnotationKey      = "lunarway.com/observed-artifact-id"
+	artifactIDAnnotationKey    = "lunarway.com/artifact-id"
+	authorAnnotationKey        = "lunarway.com/author"
+	controlledAnnotationKey    = "lunarway.com/controlled-by-release-manager"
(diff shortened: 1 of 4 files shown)


cmd/daemon/kubernetes/kubernetes.go
```diff
@@ -77,11 +77,12 @@ func isCorrectlyAnnotated(annotations map[string]string) bool {
...
-	observedAnnotationKey   = "lunarway.com/observed-artifact-id"
-	artifactIDAnnotationKey = "lunarway.com/artifact-id"
-	authorAnnotationKey     = "lunarway.com/author"
-	controlledAnnotationKey = "lunarway.com/controlled-by-release-manager"
-	squadLabelKey           = "squad"
+	observedAnnotationKey      = "lunarway.com/observed-artifact-id"
+	artifactIDAnnotationKey    = "lunarway.com/artifact-id"
+	authorAnnotationKey        = "lunarway.com/author"
+	controlledAnnotationKey    = "lunarway.com/controlled-by-release-manager"
+	runtimeAlertsAnnotationKey = "lunarway.com/runtimeAlerts"
```
Crevil, line 84: This really highlights our naming inconsistency. I guess this should be with a - and our deployments should be updated accordingly.

```suggestion
	runtimeAlertsAnnotationKey = "lunarway.com/runtime-alerts"
```

cmd/daemon/kubernetes/pods.go
```diff
@@ -274,3 +275,13 @@ func getCodeOwnerSquad(labels map[string]string) string {
 	}
 	return "no-one"
 }
+
+func shouldSquadBeAlerted(squad string, annotations map[string]string) (alertChannel string) {
+	if value, ok := annotations[runtimeAlertsAnnotationKey]; ok {
+		if value == "false" {
+			return "false"
```
Crevil, line 282: I think an empty string is better here to avoid an odd string looking like a bool value 🙈
```suggestion
			return ""
```

internal/slack/slack.go
```diff
@@ -313,6 +313,9 @@ func (c *Client) NotifyK8SPodErrorEvent(ctx context.Context, event *http.PodErro
 	if err != nil {
 		return err
 	}
+	if event.AlertSquad != "false" {
```
Crevil, line 316: Error check missing?

MariaCFFrandsen: It is a draft still :P

Crevil: Haha, oh sorry 🙈

MariaCFFrandsen: I'm not sure the error check is needed after :)
This
```
	if err != nil {
		return err
	}
	return err
```
does the same as at the end of the function 
```
return err
```

Crevil: I guess the next "feature gate" for aura and odyssey should be removed with this addition?

Currently we "could" override the `err` if a channel is set _and_ it is squad aura :)

Crevil: Just saw that it was removed 🙈

MariaCFFrandsen: It has been removed :) Then we would have our squad hardcoded into a public repo

Crevil (approved):
Just a single request left. A proper PR description on what motivates this change (the why) 🙈 👍

PR #411, opened 2022-08-27
Add release prometheus counter

Crevil:
This change registers a new release notifier which tracks a prometheus counter
for each release. The release notifier option is also expanded to include the
intent which contains relevant information on the behaviour backing the release,
eg. was it a rollback.

```
# HELP release_manager_releases_total Total number of releases
# TYPE release_manager_releases_total counter
release_manager_releases_total{environment="dev",intent="ReleaseArtifact",releaser="Bjørn Sørensen",service="test-service"} 1
```


mahlunar (approved):
🚀

PR #423, opened 2022-09-02
Inject copier instead of using globals

Crevil:
The current implementation of flows and git depends on a concrete global
implementation for copying in the copy package. This makes any usage hard to
test and also reason about. Further more does the copy functions rely on a
global logger instance making for an even more coupled implementation.

This change extracts the copying functionality into an injected struct with the
logger injected. No tests or functionality is changed. This is purely a
refactoring to enable future changes.

This is part of a larger change to remove the global logging state and inject it in all places.

### DIFF
--- a/cmd/server/command/start.go
+++ b/cmd/server/command/start.go
@@ -18,6 +18,7 @@ import (
 	"github.com/lunarway/release-manager/internal/broker"
 	"github.com/lunarway/release-manager/internal/broker/amqpextra"
 	"github.com/lunarway/release-manager/internal/broker/memory"
+	"github.com/lunarway/release-manager/internal/copy"
 	"github.com/lunarway/release-manager/internal/flow"
 	"github.com/lunarway/release-manager/internal/git"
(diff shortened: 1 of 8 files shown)


cmd/server/command/start.go
```diff
@@ -144,8 +145,12 @@ func NewStart(startOptions *startOptions) *cobra.Command {
 					}
 				}
 			}
+
+			copier := copy.New(log.With("system", "copier"))
```
mahlunar, line 149: If you want to ensure that the log system is set every time, then maybe move the call inside the constructor function?

mahlunar (approved):

cmd/server/command/start.go
Crevil: Yeah I thought we could give this some thought. I am preparing a more general PR that injects the logger everywhere and thus it becomes very clear where this might fit.

The PR just got out of hand, so I extracted some bits eg. into this change.

PR #424, opened 2022-09-04
Explain why unknown email is logged

Crevil:
This change removes a todo which questions why we log and return an error.


kjuulh (approved):
Looks good i see quite a few errors so we probably need a dashboard/view for these if possible

Crevil:
We have alerts on them and specifically in this case to follow up on 👍

PR #430, opened 2022-09-21
send release to development-metrics

MariaCFFrandsen:


internal/flow/notify.go
```diff
@@ -43,3 +50,15 @@ func (s *Service) NotifyK8SJobErrorEvent(ctx context.Context, event *http.JobErr
 	}
 	return nil
 }
+
+func notifyDevelopmentMetrics(k8sReleaseEvent *http.ReleaseEvent) error {
+	marshal, err := json.Marshal(k8sReleaseEvent)
+	if err != nil {
+		return err
+	}
+	post, err := httppackage.Post("http://development-metrics-service:3000/release-manager", "application/json", strings.NewReader(string(marshal)))
```
MariaCFFrandsen, line 59: It's probably not smart to list an internal address in an open-source project - any ideas on how to make it more generic?

MariaCFFrandsen: We are going to do have to other events as well so of course has to been nicely :)

Crevil: I actually think this should be decoupled from this service. We have a broker available so we can publish a `Released` message that can then be consumed by whom ever. This way we can add multiple listeneres in the future without having to handle this.

This also decouples the services as no run-time requirements for their existence is needed. WDYT?

MariaCFFrandsen: That is 100% a better approach. We have started out with a rest api on the development-metrics service but hopefully we can switch fast!

I'm assuming we still want to publish the Release event here. Should we add a Broker to service called here?

Crevil: Yeah, I'm actually not sure if we already publish some of these actually. 🤔 But if not, I think we should focus on that instead. Maybe @mahlunar or @jonasvinther can help out on this part if you want to divide and conquer so they add publishing to release-manager and you can focus on consuming.

MariaCFFrandsen: That would be awesome!

As far as I can, it only publishes service and artifact id atm. 
```flowSvc.PublishReleaseArtifactID = func(ctx context.Context, event flow.ReleaseArtifactIDEvent) error {
				return brokerImpl.Publish(ctx, &event)
			}
			flowSvc.PublishNewArtifact = func(ctx context.Context, event flow.NewArtifactEvent) error {
				return brokerImpl.Publish(ctx, &event)
			}```
 in start.go with

```
type NewArtifactEvent struct {
	Service    string `json:"service,omitempty"`
	ArtifactID string `json:"artifactId,omitempty"`
}
```
I dont think we can utilize these instead of the full release event.

Comment (Crevil) [reply internal/flow/notify.go:59]: No that is true. Another benefit of getting the "It happened" messages published is we can collect it into the Lunar data system which makes it possible to dashboard over long periodes 😄

Comment (MariaCFFrandsen) [reply internal/flow/notify.go:59]: That would be awesome!

As far as I can, it only publishes service and artifact id atm. 
```flowSvc.PublishReleaseArtifactID = func(ctx context.Context, event flow.ReleaseArtifactIDEvent) error {
		return brokerImpl.Publish(ctx, &event)
	}
    flowSvc.PublishNewArtifact = func(ctx context.Context, event flow.NewArtifactEvent) error {
		return brokerImpl.Publish(ctx, &event)
	}```


 in start.go and event:

We can use this event 
```
type ReleaseArtifactIDEvent struct {
	Service     string        `json:"service,omitempty"`
	Environment string        `json:"environment,omitempty"`
	Namespace   string        `json:"namespace,omitempty"`
	ArtifactID  string        `json:"artifactID,omitempty"`
	Branch      string        `json:"branch,omitempty"`
	Actor       Actor         `json:"actor,omitempty"`
	Intent      intent.Intent `json:"intent,omitempty"`
}
```

Comment (MariaCFFrandsen) [reply internal/flow/notify.go:59]: I'm sure what trigger it though

**Comment by MariaCFFrandsen** [comment]
rabbitmq messages wish list 

type ReleaseEvent struct { 
	Name          string `json:"name,omitempty"`
	Namespace     string `json:"namespace,omitempty"`
	ResourceType  string `json:"resourceType,omitempty"`
	AvailablePods int32  `json:"availablePods"`
	DesiredPods   int32  `json:"replicas,omitempty"`
	ArtifactID    string `json:"artifactId,omitempty"`
	AuthorEmail   string `json:"authorEmail,omitempty"`
	Environment   string `json:"environment,omitempty"`
} Squad should ideally be included, it would be awesome.

Instead of three events for each k8s fail error, it would be awesome with a fail reason/type
type K8sFailEven struct { 
	Name          string `json:"name,omitempty"`
	Namespace     string `json:"namespace,omitempty"`
	ResourceType  string `json:"resourceType,omitempty"`
	AvailablePods int32  `json:"availablePods"`
	DesiredPods   int32  `json:"replicas,omitempty"`
	ArtifactID    string `json:"artifactId,omitempty"`
	AuthorEmail   string `json:"authorEmail,omitempty"`
	Environment   string `json:"environment,omitempty"`
} 

type ContainerError struct {
	Name         string `json:"name,omitempty"`
	ErrorMessage string `json:"message,omitempty"`
	Type         string `json:"type,omitempty"`
	ArtifactID    string `json:"artifactId,omitempty"` //should be added
}

I am fairly certain we need ArtifactID to correlate events.

File: internal/flow/notify.go [hunk 2]
```diff
@@ -13,6 +16,10 @@ func (s *Service) NotifyK8SDeployEvent(ctx context.Context, event *http.ReleaseE
 	defer span.Finish()
 	span, _ = s.Tracer.FromCtx(ctx, "post k8s deploy slack message")
 	err := s.Slack.NotifyK8SDeployEvent(ctx, event)
+	err = notifyDevelopmentMetrics(event)
```
kjuulh, line 19: This should not be within the tracer span.

internal/flow/notify.go
```diff
@@ -13,6 +16,10 @@ func (s *Service) NotifyK8SDeployEvent(ctx context.Context, event *http.ReleaseE
...
+	if err != nil {
+		errors.WithMessage(err, "post k8s deploy development metrics")
```
kjuulh, line 21: missing return

PR #431, opened 2022-09-21
Feature/add release event publishing

kjuulh:
This is the first step of adding events to be processed both internally and externally.

We'll start with release event

@MariaCFFrandsen maybe you would like to comment on the model, some of the fields you mentioned, sadly isn't possible. But I've taken the liberty to add a few more.

### DIFF
--- a/cmd/server/command/start.go
+++ b/cmd/server/command/start.go
@@ -19,6 +19,7 @@ import (
 	"github.com/lunarway/release-manager/internal/broker/amqpextra"
 	"github.com/lunarway/release-manager/internal/broker/memory"
 	"github.com/lunarway/release-manager/internal/copy"
+	"github.com/lunarway/release-manager/internal/events"
 	"github.com/lunarway/release-manager/internal/flow"
 	"github.com/lunarway/release-manager/internal/git"
 	"github.com/lunarway/release-manager/internal/github"
@@ -46,6 +47,7 @@ const (
 func (t *brokerType) String() string {
 	return string(*t)
 }
+
 func (t *brokerType) Set(s string) error {
 	switch brokerType(s) {
 	case BrokerTypeMemory, BrokerTypeAMQP:
@@ -55,6 +57,7 @@ func (t *brokerType) Set(s string) error {
 		return errors.New("broker not supported")
 	}
 }
+
 func (t *brokerType) Type() string {
 	return "string"
 }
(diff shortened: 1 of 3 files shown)


cmd/server/command/start.go
```diff
@@ -249,37 +280,65 @@ func NewStart(startOptions *startOptions) *cobra.Command {
...
+					metricsObserver.ObserveRelease(
+						metrics.Release{
+							Environment: opts.Environment,
+							Service:     opts.Service,
+							Releaser:    opts.Releaser,
+							Intent:      opts.Intent.Type,
+						},
+					)
+				},
+				"events": func(ctx context.Context, opts flow.NotifyReleaseOptions) {
```
MariaCFFrandsen, line 304: @kjuulh Looks good to me - Thanks for spending time on it!

I have some questions:)
It means that the queue defaults to "release-manager" - do you know if this is true? And do you know where the value is retrieved from on start https://github.com/lunarway/release-manager/blob/master/cmd/server/command/root.go#L129?

release-manager uses the github.com/rabbitmq/amqp091-go directly. 
When we need to consume, would be possible to go through go.lunarway.com/amqp and what approach would you recommend?

kjuulh: We cannot go through go.lunarway.com/amqp as that is a private library. You would need to listen on the exchange in the same manner as we do in release manager it self

MariaCFFrandsen: No, but we can in development-metrics :) Alright.

cmd/server/command/start.go
```diff
@@ -249,37 +280,65 @@ func NewStart(startOptions *startOptions) *cobra.Command {
...
 						Service:     opts.Service,
-						Releaser:    opts.Releaser,
-						Intent:      opts.Intent.Type,
+						Namespace:   opts.Namespace,
+						ArtifactID:  opts.Spec.ID,
+						AuthorEmail: opts.Spec.Application.AuthorEmail,
+						AuthorName:  opts.Spec.Application.AuthorName,
+						Environment: opts.Environment,
+						Squad:       opts.Spec.Squad,
+						IntentType:  opts.Intent.Type,
```
Crevil, line 314: I think some of the intent details are valuable to provide as well, eg. if intent is `Rollback` the `PreviousArtifactID` is nice to know "from where" the rollback happened.

@MariaCFFrandsen has some ideas here that we will flesh out tomorrow in a DDD session.

cmd/server/command/start.go
```diff
@@ -249,37 +280,65 @@ func NewStart(startOptions *startOptions) *cobra.Command {
...
 					})
+
+					if err != nil {
+						logger.Errorf("could not publish released_event. Error: %w", err)
+					} else {
+						logger.Debugf("published released_event for artifactId: %s", opts.Spec.ID)
+					}
```
Crevil, line 321: We could drop the `else` here as we tend to in Go 🙈 

```suggestion
						return
					}
					logger.Debugf("published released_event for artifactId: %s", opts.Spec.ID)
```

cmd/server/command/start.go
```diff
@@ -309,26 +367,40 @@ func NewStart(startOptions *startOptions) *cobra.Command {
 					}
 					return flowSvc.ExecNewArtifact(context.Background(), event)
 				},
+				events.ReleasedEvent{}.Type(): func(d []byte) error {
+					var event events.ReleasedEvent
```
Crevil, line 371: Do we need to add a handler for it?

MariaCFFrandsen (approved):

cmd/server/command/start.go
kjuulh: I will add poderror event when we've gotten your subscription stuff to work

Crevil (approved):
Overall I think this is sound. There are _a lot_ of unrelated changes in the PR which makes it hard to read. I guess the functional change is really just half the changed lines if not less 🙈

kjuulh:
Yeah much less we probably need a full gofmt pr just so that we don't have to bother each following pr

PR #436, opened 2022-10-03
Revert 'remove-duplicate-branch'

kjuulh:
We've seen issues with v0.24 where we cannot release artifacts from the cli, we suspect it is because of this

### DIFF
--- a/cmd/hamctl/command/actions/artifact_id.go
+++ b/cmd/hamctl/command/actions/artifact_id.go
@@ -37,7 +37,7 @@ func ArtifactIDFromBranch(client *httpinternal.Client, service string, branch st
 	var describeResp httpinternal.DescribeArtifactResponse
 	params := url.Values{}
 	params.Add("branch", branch)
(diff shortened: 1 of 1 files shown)


MariaCFFrandsen (approved):

Crevil:
I actually think the issue might be with #407 instead. The reverted change here was found _after_ 407 was merged as promotes didn't work and reading the code nothing is using this.

We could consider making the new mux more forgiving for this to avoid the hamctl version breaking ?

kjuulh:
> @Crevil 

Lets do that. I will hold for now, but release the current version. If instead it breaks again after updating the hamctl, then lets consider that, Thoughts?

Crevil:
Yeah. Upgrading hamctl is done in https://github.com/lunarway/lw-zsh-versions/blob/master/binary_versions which should be aligned with the release internally 👍

kjuulh:
> 

Ah so we actually need to update that one as well 🤦

PR #444, opened 2022-10-16
Publish ReleaseSucceeded event when daemon reports success

Crevil:
This change adds a new integration event indicating that a daemon has observed a successful release of an artifact.

The flow service method `NotifyK8SDeployEvent` is changed to trigger hooks like `NotifyReleaseHook` and the slack message is extracted into `start.go` along with the new event hook.

The payload is currently not stable and just contains the same fields as the HTTP payload.


internal/events/release_succeeded.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// TODO: Make generic when go1.18 is adopted
+// Marshal implements broker.Publishable
+func (re ReleaseSucceeded) Marshal() ([]byte, error) {
+	return json.Marshal(re)
+}
+
+// Unmarshal implements broker.Publishable
+func (re ReleaseSucceeded) Unmarshal(data []byte) error {
```
kjuulh, line 27: don't you want pointer-receiver here?

internal/events/release_succeeded.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/lunarway/release-manager/internal/broker"
+)
+
+type ReleaseSucceeded struct {
+	Name          string `json:"name,omitempty"`
+	Namespace     string `json:"namespace,omitempty"`
+	ResourceType  string `json:"resourceType,omitempty"`
+	AvailablePods int32  `json:"availablePods,omitempty"`
+	DesiredPods   int32  `json:"desiredPods,omitempty"`
+	ArtifactID    string `json:"artifactID,omitempty"`
```
kjuulh, line 15: is artifactID also capitalized `ID` in the json as well?

kjuulh (commented):
nits

internal/events/release_succeeded.go
Crevil: This is the same with the other events. I think going for a generics implementation as noted elsewhere might be a way forward on this.

internal/events/release_succeeded.go
Crevil: Great catch 🤚

kjuulh (approved):

PR #450, opened 2022-10-24
Make sure we don't crash on missing email

kjuulh:
If we cannot find the .gitconfig/comitter then we get a panic because the `committer` object is nil. As such we now use an else statement and report error in the case it is missing.

Fixes: AURA-147

### DIFF
--- a/cmd/hamctl/command/root.go
+++ b/cmd/hamctl/command/root.go
@@ -58,8 +58,9 @@ func NewRoot(version *string) (*cobra.Command, error) {
 				committer, err := gitConfigAPI.CommitterDetails()
 				if err != nil {
 					missingFlags = append(missingFlags, "user-email")
+				} else {
+					email = committer.Email
 				}
-				email = committer.Email
 			}
 			client.Metadata.CallerEmail = email
 			if len(missingFlags) != 0 {
--- a/cmd/hamctl/command/root.go
+++ b/cmd/hamctl/command/root.go
@@ -53,7 +53,7 @@ func NewRoot(version *string) (*cobra.Command, error) {
 			if service == "" {
 				missingFlags = append(missingFlags, "service")
 			}
-			if err := setCallerEmailFromCommitter(gitConfigAPI, &client, email); err != nil {
+			if err := setCallerEmail(gitConfigAPI, &client, email); err != nil {
 				missingFlags = append(missingFlags, "user-email")
 			}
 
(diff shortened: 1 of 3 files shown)


mahlunar (commented):
Please add a test as well 👍🏼

cmd/hamctl/command/root.go
```diff
@@ -58,8 +58,9 @@ func NewRoot(version *string) (*cobra.Command, error) {
 				committer, err := gitConfigAPI.CommitterDetails()
 				if err != nil {
 					missingFlags = append(missingFlags, "user-email")
+				} else {
```
Crevil, line 61: This really feels wrong. If `err != nil` I would expect `comitter` to at least be defined. Always.
Isn't this the API of `gitConfigAPI` that is not behaving well?

kjuulh: I don't really agree. Right now in the code, we don't exit even if we have an error (until further down the line), however, we still depend on the committer struct being present, which in this case it isn't because it is a reference. We could change it to a struct, but it feels wrong to depend on default values on an object that shouldn't be set (if an error occurs).

I.e. if err is defined we only trust the error value and nothing else. 
If the err is not defined we can trust committer, because that is the api.

Also the only reason why we need the else, is because something further down are bundling up these missing flags and throwing an error.

Crevil: I see 👍 Makes fine sense. I misread the `if` condition here and expected "it handles the error case" where in fact it just collects flag error cases and not flow control 🙈

kjuulh:
@mahlunar and @Crevil I've added tests and refactored a bit.

cmd/hamctl/command/root.go
```diff
@@ -54,14 +53,10 @@ func NewRoot(version *string) (*cobra.Command, error) {
 			if service == "" {
 				missingFlags = append(missingFlags, "service")
 			}
-			if email == "" {
-				committer, err := gitConfigAPI.CommitterDetails()
-				if err != nil {
-					missingFlags = append(missingFlags, "user-email")
-				}
-				email = committer.Email
+			if err := setCallerEmailFromCommitter(gitConfigAPI, &client); err != nil {
```
Crevil, line 56: This doesn't respect setting the `user-email` flag directly, does it?
In that case `email` will be set up front?

Specifically this is a change as the `if email == ""` is removed?

kjuulh: I've included the args as well

Crevil (approved):

PR #457, opened 2022-11-14
release-manager.#.messagetype

MariaCFFrandsen:


kjuulh (approved):
@Crevil do you see anything out of place? I think we should just give it a shot. As far as I can tell only the new events actually use the router.

internal/broker/amqpextra/publish.go
```diff
@@ -19,7 +20,7 @@ func (w *Worker) Publish(ctx context.Context, message broker.Publishable) error
 
 	err := w.worker.Publish(ctx, amqp.PublishDto{
 		Exchange:      w.config.Exchange,
-		RoutingKey:    w.config.RoutingKey,
+		RoutingKey:    strings.Join([]string{w.config.RoutingKey, message.Type()}, "."),
```
Crevil, line 23: One couæd argue this is a complex way for formatting a string 🙈

```suggestion
		RoutingKey:    fmt.Sprintf("%s.%s", w.config.RoutingKey, message.Type()),
```

Crevil (commented):
We subscribe to messages put selves right? So do the consumer need a change as well?

MariaCFFrandsen:
> We subscribe to messages put selves right? So do the consumer need a change as well?

I would accept the tests to enlighten this?

PR #477, opened 2023-03-27
Route all messages from the release-manager exchange

mahlunar:
Currently, the queue consumer will not receive all messages as it only has a binding on a hard-coded value. This change allows is to listed for all messages from its exchange.


kjuulh (approved):
damn. forgot to do it friday... 

🙈

PR #498, opened 2023-11-05
Oauth2 support for release manager

hoeg:
This PR contains the introduction of the `login` command to `hamctl` which will initiate an oauth2 device authorization flow.

The access token is persisted to disk and when it is expired the user must do a login again.

The release daemon and artifact executables has been moved to client credentials flow to have oauth2 on all the calls.

We support both the old auth token that is shared as well as the new jwt. The server must be configured with an IdP but this version is kept backwards compatible such that old clients can still use the server. When users have updated their cli's then we can remove the support for the static token.

The changes include that:
 
- We require oauth2 configuration to be set in the environment.
- The access token is persisted to disk in the `$HOME/.Config` directory (should be ok on all OS, right?)
- We bumped the golang version to v 1.20 to have access to the `golang.org/x/oauth2` package
- Support both the old Authorization token and the new jwt in the server. The `Actor` is taken from the context, in case of the jwt supplied for auth or from the request if the old static access token is used.

We have been holding back on documentation as we would like to agree to the changes before writing it 🚀


cmd/hamctl/command/root.go
```diff
@@ -34,7 +45,7 @@ func NewRoot(version *string) (*cobra.Command, error) {
 		PersistentPreRunE: func(c *cobra.Command, args []string) error {
 			// all commands but version and completion requires the "service" flag
 			// if this is one of them, skip the check
```
Crevil, line 47: Classic comment issue. It tells what the code does, and now the code changed 🙈🤥

I guess we should remove it or update it.

cmd/hamctl/command/root.go
```diff
@@ -45,17 +56,10 @@ func NewRoot(version *string) (*cobra.Command, error) {
...
 
-			if client.Metadata.AuthToken == "" {
-				client.Metadata.AuthToken = os.Getenv("HAMCTL_AUTH_TOKEN")
-			}
-
 			var missingFlags []string
 			if service == "" {
 				missingFlags = append(missingFlags, "service")
 			}
-			if err := setCallerEmail(gitConfigAPI, &client, email); err != nil {
```
Crevil, line 56: Doesn't this make the gitConfigAPI unused and maybe even the whole local git config functionality could be removed? 🥳

internal/http/authenticator.go
```diff
@@ -0,0 +1,120 @@
...
+		},
+		Scopes: []string{"openid profile"},
+	}
+	return UserAuthenticator{
+		conf: conf,
+	}
+}
+
+func (g *UserAuthenticator) Login() error {
+	ctx := context.Background()
```
Crevil, line 35: Maybe we should accept the context as an argument so the cobra CLI can cancel it correctly?

internal/http/client.go
```diff
@@ -72,9 +79,8 @@ func (c *Client) Do(method string, path string, requestBody, responseBody interf
 	if err == nil {
 		req.Header.Set("x-request-id", id.String())
 	}
-	req.Header.Set("Authorization", "Bearer "+c.Metadata.AuthToken)
 	req.Header.Set("X-Cli-Version", c.Metadata.CLIVersion)
-	req.Header.Set("X-Caller-Email", c.Metadata.CallerEmail)
+	req.Header.Set("X-HAM-TOKEN", "test")
```
Crevil, line 83: Is this intentional?

Crevil (commented):
Awesome to see this addition! Did you look into what it would take to verify it on the server as well?

I know you wrote about the proxy in the PR description, I'm just curious as to what it would look like 🤘

hoeg:
> Awesome to see this addition! Did you look into what it would take to verify it on the server as well?
> 
> I know you wrote about the proxy in the PR description, I'm just curious as to what it would look like 🤘

Its a matter of changing the middleware and provide the idp config on startup. Seems pretty simple, should i take a stab at adding that part and Remote the ham token all togethers?

cmd/artifact/command/push.go
```diff
@@ -24,12 +25,26 @@ func pushCommand(options *Options) *cobra.Command {
 			var err error
 			ctx := context.Background()
 
-			if releaseManagerClient.Metadata.AuthToken != "" {
```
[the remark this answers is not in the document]
hoeg: It is, i cannot see the reason for it not being set. Please enlighten me on the usecase 🙏

internal/http/client.go
hoeg: Yes, Im an still in draft, so i figures i would just make tests pass and then get some feedback regarding the ude of this token (as mentioned in the PR) before we do something about it ✌️ this Will most definetely make it to the final change set 😅

Crevil:
Depends on how much it ends up to be I guess. Aura aren’t super comfortable with the codebase but if it isn’t a large change it might be simple enough to incorporate🙏

hoeg:
implemented token verification in the server as well instead of relying on a proxy for handling the token.

cmd/hamctl/command/root.go
```diff
@@ -72,18 +72,17 @@ func NewRoot(version *string) (*cobra.Command, error) {
...
 		NewPromote(&client, &service, releaseClient),
 		NewRelease(&client, &service, loggerFunc, releaseClient),
 		NewRollback(&client, &service, loggerFunc, SelectRollbackReleaseFunc, releaseClient),
 		NewStatus(&client, &service),
 		NewVersion(*version),
+		Login(authenticator),
 	)
 	command.PersistentFlags().DurationVar(&client.Timeout, "http-timeout", 120*time.Second, "HTTP request timeout")
 	command.PersistentFlags().StringVar(&client.BaseURL, "http-base-url", "", "address of the http release manager server")
-	command.PersistentFlags().StringVar(&client.Metadata.AuthToken, "http-auth-token", "", "auth token for the http service")
```
Crevil, line 84: To make a gradual roll out we might consider only deprecating the flags and remove them in a later release? That way we can update any use of them without everything breaking.

Just a thought 👌

cmd/server/http/jwt.go
```diff
@@ -0,0 +1,134 @@
...
+
+// authenticate authenticates the handler against a Bearer token.
+//
+// If authentication fails a 401 Unauthorized HTTP status is returned with an
+// ErrorResponse body.
+func (v *Verifier) authentication(token string) func(http.Handler) http.Handler {
+	return func(h http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			bearerToken, err := ParseBearerToken(token)
+			if err != nil {
```
Crevil, line 73: This conditional is a bit hard to grok. If the server is started with a JWT as hamctl token, it will parse the request's JWT? Or? No, that is not right. The `token` variable is from the server startup arguments. 🧐

What is going on here?
Maybe it should be extracted into some named functions to make it more clear what happens in each clause?

cmd/server/http/jwt.go
```diff
@@ -0,0 +1,134 @@
...
+				authorization := r.Header.Get("Authorization")
+				t := strings.TrimPrefix(authorization, "Bearer ")
+				t = strings.TrimSpace(t)
+				if t != token {
+					httpinternal.Error(w, "please provide a valid authentication token", http.StatusUnauthorized)
+					return
+				}
+			} else {
+				keySet, err := v.jwkCache.Get(context.Background(), v.jwksLocation)
```
Crevil, line 82: The request's context should be used here, right?

cmd/server/http/jwt.go
```diff
@@ -0,0 +1,134 @@
...
+				if err != nil {
+					httpinternal.Error(w, "please provide a valid authentication token", http.StatusUnauthorized)
+					return
+				}
+
+				parsedToken, err := v.verify(bearerToken, keySet)
+				if err != nil {
+					if strings.Contains(err.Error(), keyNotFoundMsg) {
+						freshKeys, err := v.jwkCache.Refresh(context.Background(), v.jwksLocation)
```
Crevil, line 91: Same regarding the context as above?

cmd/server/http/jwt.go
```diff
@@ -0,0 +1,134 @@
...
+						if err != nil {
+							httpinternal.Error(w, "please provide a valid authentication token", http.StatusUnauthorized)
+							return
+						}
+					} else {
+						httpinternal.Error(w, "please provide a valid authentication token", http.StatusUnauthorized)
+						return
+					}
+				}
+				context.WithValue(r.Context(), AUTH_USER_KEY, parsedToken.Subject())
```
Crevil, line 106: The return value is not assigned here? It should replace the request's context, right?

cmd/server/http/policy.go
```diff
@@ -30,12 +30,19 @@ func applyAutoReleasePolicy(payload *payload, policySvc *policyinternal.Service)
...
 
-		logger = logger.WithFields("service", req.Service, "req", req)
-		logger.Infof("http: policy: apply: service '%s' branch '%s' environment '%s': apply auto-release policy started", req.Service, req.Branch, req.Environment)
-		id, err := policySvc.ApplyAutoRelease(ctx, policyinternal.Actor{
+		actor := policyinternal.Actor{
 			Name:  req.CommitterName,
 			Email: req.CommitterEmail,
-		}, req.Service, req.Branch, req.Environment)
+		}
+		subject := r.Context().Value(AUTH_USER_KEY).(string)
```
Crevil, line 37: This might be a good candiate for a `UserKeyFromContect()` function to hide the nasty lookup/casting stuff?

Crevil (commented):
Whoa, that wasn't too bad IMO 🤘

Have you had it running locally and seen it work? 🥳
Otherwise I might take a stab at it tonight 🤘

hoeg:
> Whoa, that wasn't too bad IMO 🤘
> 
> Have you had it running locally and seen it work? 🥳 Otherwise I might take a stab at it tonight 🤘

I have not had the time to set it up locally. Please go for it if you have the time, if not then I will get to it later this week.

Also, if you find anything during your test you are of course more then welcome to push changes to this branch 🙏

hoeg:
Converted back to draft to ensure we iron the last parts before Aura takes a look

cmd/artifact/command/push.go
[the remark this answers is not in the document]
hoeg: @aura what do you think about this?

cmd/server/http/jwt.go
```diff
@@ -0,0 +1,151 @@
...
+	tokenParts := strings.Split(jwt, ".")
+
+	if len(tokenParts) != 3 {
+		return "", errors.New("invalid token format")
+	}
+
+	return strings.TrimSpace(jwt), nil
+}
+
+const AUTH_USER_KEY = "AUTH_USER_KEY"
```
Crevil, line 64: A best practice is to have a complex type to avoid anyone outside the package to access the subject.

```suggestion
type AUTH_USER_KEY struct{}
```

cmd/daemon/command/start.go
```diff
@@ -34,6 +34,22 @@ func StartDaemon() *cobra.Command {
 			logConfiguration.ParseFromEnvironmnet()
 			log.Init(logConfiguration)
 
+			idpURL := os.Getenv("HAMCTL_OAUTH_IDP_URL")
```
Crevil, line 37: Per conversation: This should be flags instead and the hamctl part should be removed.

cmd/artifact/command/push.go
```diff
@@ -24,12 +25,26 @@ func pushCommand(options *Options) *cobra.Command {
...
-				artifactID, err = flow.PushArtifactToReleaseManager(ctx, &releaseManagerClient, options.FileName, options.RootPath)
-				if err != nil {
-					return err
-				}
+			idpURL := os.Getenv("HAMCTL_OAUTH_IDP_URL")
```
Crevil, line 28: Per conversation: this should be flags and hamctl prefix should be removed

cmd/hamctl/command/root.go
Crevil: I just checked and this specific flag isn't used. We should be all good. The server middleware supports both authentication mechanisms, so users with an older hamctl will still have access during the rollout.

cmd/artifact/command/push.go
hoeg: Thinking about it, I don't like to give the client secret as a command line argument as that leaks out to e.g. /proc/pid/cmdline. I will put the secret in a file and take an argument with a path to that file 🚀

Crevil: As all secrets are currently mounted as args I would argue we should follow that convention for now.

We can then iterate on a follow up changing all of them in one go.

hoeg:
@Crevil I think we have been through all the comments now. I guess we just have to test the daemon and artifact cli's before going out of draft, right?

hoeg:
Should we initiate a login if the token is invalid or should we just fail and tell the user to do a login instead?

Doing automatic logins will hopefully make for at better user experience but that might break e.g. scripts that cannot pop a browser.
@AndersSoee suggested that we could add an additional flag that indicates that we should not auto-login if the token is valid.

Currently we just fail if the token is invalid and tell the user to log in.

cmd/server/http/jwt.go
```diff
@@ -0,0 +1,163 @@
...
+	value := ctx.Value(authenticatedUserContextKey{})
+
+	if value == nil {
+		return ""
+	}
+
+	return value.(string)
+}
+
+func NewVerifier(ctx context.Context, jwksLocation string, jwkFetchTimeout time.Duration, issuer string, audience string) (*Verifier, error) {
```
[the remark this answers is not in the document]
hoeg: Should probably be configurable and set with a default value?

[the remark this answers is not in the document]
Crevil: Yeah maybe. But what is the purpose of it? It isn't used, so is some logic missing?
As it is a timeout I wouldn't expect configuration of it to be super important as those kind of values tend to be pretty static (compared to lifetimes, etc.)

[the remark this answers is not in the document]
hoeg: lets just remove it then. Might be a timeout that is not yet set on the http client that is used to fetch the jwks. I am unsure if we would ever change it so might be a good idea just to remove it and keep the defailt http timeout.

cmd/server/http/jwt.go
```diff
@@ -0,0 +1,163 @@
...
+		jwkCache:     cache,
+		issuer:       issuer,
+		audience:     audience,
+	}, nil
+}
+
+func ParseBearerToken(token string) (string, error) {
+	jwt := strings.TrimPrefix(token, "Bearer")
+
+	tokenParts := strings.Split(jwt, ".")
```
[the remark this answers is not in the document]
hoeg: Pushed a fix 🚀

hoeg:
Guess we are ready for review now @Crevil?

Crevil:
Yeah, I think so. Should we do a fresh PR to reduce the overhead of all our previous communication?

hoeg:
Sure, I will close this PR and create a new one and make sure to add the outstanding questions to the description 🚀

PR #502, opened 2023-11-13
Add OAuth2 support

hoeg:
This PR contains the addition of the login command to hamctl which will initiate an oauth2 device authorization flow.

The access token is persisted to disk and when it is expired the user must do a login again.

The release daemon and artifact executables has been moved to client credentials flow to have oauth2 on all the calls.

We support both the old auth token that is shared as well as the new jwt. The server must be configured with an IdP but this version is kept backwards compatible such that old clients can still use the server. When users have updated their cli's then we can remove the support for the static token.

The changes include that:

- We require oauth2 configuration to be set in the environment.
- The access token is persisted to disk in the $HOME/.Config directory (should be ok on all OS, right?)
- We bumped the golang version to v 1.20 to have access to the golang.org/x/oauth2 package
- Support both the old Authorization token and the new jwt in the server. The Actor is taken from the context, in case of the jwt supplied for auth or from the request if the old static access token is used.

Question for the reviewers:

Should we initiate a login if the token is invalid or should we just fail and tell the user to do a login instead?

Doing automatic logins will hopefully make for at better user experience but that might break e.g. scripts that cannot pop a browser.
@AndersSoee suggested that we could add an additional flag that indicates that we should not auto-login if the token is invalid.

Currently we just fail if the token is invalid and tell the user to log in.

We have been holding back on documentation as we would like to agree to the changes before writing it 🚀


cmd/artifact/command/push.go
```diff
@@ -15,21 +15,22 @@ import (
...
 		Use:   "push",
 		Short: "push artifact to artifact repository",
 		RunE: func(c *cobra.Command, args []string) error {
 			var artifactID string
 			var err error
 			ctx := context.Background()
+			authenticator := httpinternal.NewClientAuthenticator(clientID, clientSecret, idpURL)
+			releaseManagerClient.Auth = &authenticator
 
-			if releaseManagerClient.Metadata.AuthToken != "" {
```
hoeg, line 27: Notice the change in behaviour here. We cannot see a reason to start up the artifact call without credentials. Therefore we do not support it anymore. If this is wrong, please let us know.

cmd/server/http/http.go
```diff
@@ -140,22 +138,3 @@ func trace(tracer tracing.Tracer) func(http.Handler) http.Handler {
 		})
 	}
 }
-
-// authenticate authenticates the handler against a Bearer token.
-//
-// If authentication fails a 401 Unauthorized HTTP status is returned with an
-// ErrorResponse body.
-func authenticate(token string) func(http.Handler) http.Handler {
```
mahlunar, line 148: It would be great to allow both authentication mechanisms for a while to test the new login flow before forcing everyone to use it.

mahlunar (commented):
Awesome.

I have som clarifying questions.

For us to release this gradually, then we should allow both authentication strategies. We could control this via a flag to avoid coupling a release here with out own deployment.

A release would follow these steps:

1. Deploy release manger with support for IDP
2. Move beta tests to use the IDP while the rest uses the old authentication flow
3. Move all developers to IDP flow via new hamctl version
4. Remove support for old authentication flow.

Can should be able to do the above steps after this PR is merged.

hoeg:
> Awesome.
> 
> I have som clarifying questions.
> 
> For us to release this gradually, then we should allow both authentication strategies. We could control this via a flag to avoid coupling a release here with out own deployment.
> 
> A release would follow these steps:
> 
> 1. Deploy release manger with support for IDP
> 2. Move beta tests to use the IDP while the rest uses the old authentication flow
> 3. Move all developers to IDP flow via new hamctl version
> 4. Remove support for old authentication flow.
> 
> Can should be able to do the above steps after this PR is merged.

That is exactly how we also think the release procedure should be. We do allow both mechanisms so it is doable. I will address the other comments and resolve when they are done. 

Next step is to figure out the IDP configuration, we do have a test setup that we have used to verify this implementation, but we should have something production ready set up.

I will let you know when we have resolved all comments 🚀

internal/http/client.go
```diff
@@ -52,9 +56,12 @@ func (c *Client) URLWithQuery(path string, queryParams url.Values) (string, erro
 // the server returns a status code above 399 the response is parsed as an
 // ErrorResponse object and returned as the error.
 func (c *Client) Do(method string, path string, requestBody, responseBody interface{}) error {
-	client := &http.Client{
-		Timeout: c.Timeout,
+	ctx := context.Background()
```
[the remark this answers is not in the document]
hoeg: Can we do this as a separate issue? Requires some more wiring and is not straight forward to do.

mahlunar (approved):

PR #519, opened 2024-01-18
support auto login

hoeg:
Adds the env var `HAMCTL_OAUTH_AUTO_LOGIN`. It defaults to true but you can opt out of the auto login.

We initiate a login if the token is invalid and we have configured auto login.


mahlunar (commented):
This is awesome. Let's add some docs so that it is clear how to use it.

mahlunar (approved):

PR #522, opened 2024-01-19
Add auth tests

hoeg:
Restructured for testability and added some tests for auto login

### DIFF
--- a/internal/http/authenticator_test.go
+++ b/internal/http/authenticator_test.go
@@ -5,11 +5,12 @@ import (
 	"fmt"
 	"net/http"
 	"testing"
+	"time"
 
 	"golang.org/x/oauth2"
 )
 
-func TestAccessCallsLoginAndStoresToken(t *testing.T) {
+func TestAccessNoTokenCallsLoginAndStoresToken(t *testing.T) {
 	httpAuth := HttpAuth{
 		DeviceResp: &oauth2.DeviceAuthResponse{
 			UserCode:                "YOLO",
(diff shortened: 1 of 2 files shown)


internal/http/authenticator_test.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	ua := UserAuthenticator{
+		conf:       &httpAuth,
+		ts:         &dummyStore,
+		autoLogin:  true,
+		popBrowser: false,
+	}
+
+	client, err := ua.Access(context.Background())
+	if err != nil {
```
kjuulh, line 31: I'd recommend to use require.NoError instead with a msg if wanted.

internal/http/authenticator_test.go
```diff
@@ -0,0 +1,152 @@
...
+		t.Fatal(err)
+	}
+	if client == nil {
```
kjuulh, line 34: and assert.NotNil etc for the tests below

internal/http/authenticator_test.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	if dummyStore.tokenStored {
+		t.Fatal("did not expect the token to be stored")
+	}
+	if httpAuth.loginCalled {
+		t.Fatal("did not expect login to be called")
+	}
+}
+
+type HttpAuth struct {
```
kjuulh, line 117: is there a reason you're using a mocking framework?

kjuulh: If you prefer this approach then I am fine xD

kjuulh (commented):
Some small nits

PR #531, opened 2024-03-14
Fix OOMKilled check and update OOMKilled log

JacobValdemar:
The current check was not correct. In many cases it didn't catch pods that was OOM killed. This is corroborated by https://github.com/xing/kubernetes-oom-event-generator/blob/master/src/controller/controller.go#L150. Further details provided in Slack.

### DIFF
--- a/cmd/daemon/kubernetes/pods.go
+++ b/cmd/daemon/kubernetes/pods.go
@@ -128,7 +128,7 @@ func (p *PodInformer) handle(e interface{}) {
 	}
 
 	if isPodOOMKilled(pod) {
-		log.With("killedPodNamespace", pod.Namespace).Infof("Pod: %s was OOMKilled owned by squad %s", pod.Name, getCodeOwnerSquad(pod.Labels))
+		log.With("targetPodNamespace", pod.Namespace, "targetPodName", pod.Name, "targetPodSquad", getCodeOwnerSquad(pod.Labels)).Infof("Pod: %s was OOMKilled owned by squad %s", pod.Name, getCodeOwnerSquad(pod.Labels))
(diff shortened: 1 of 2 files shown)


cmd/daemon/kubernetes/pods.go
```diff
@@ -128,7 +128,7 @@ func (p *PodInformer) handle(e interface{}) {
 	}
 
 	if isPodOOMKilled(pod) {
-		log.Infof("Pod: %s was OOMKilled owned by squad %s", pod.Name, getCodeOwnerSquad(pod.Labels))
+		log.With("killedPodNamespace", pod.Namespace).Infof("Pod: %s was OOMKilled owned by squad %s", pod.Name, getCodeOwnerSquad(pod.Labels))
```
kjuulh, line 131: this seems like a strange name for a label?

Or we can get a little more generic name for the pod or put it in the template itself.

We could just as well serialize all the information about the pod and put that in the log as well, then we can extract the parts we need.

JacobValdemar: @kjuulh I am open to suggestions for a different field name, as long as it doesn't collide with existing labels. I don't think it makes sense to log everything (you have the final say as you own the service)

kjuulh: Maybe we could call it `targetPodNamespace`, we could also move pod down so that it would be called `targetPodName`.

JacobValdemar: @kjuulh What do you think about b8df41c?

kjuulh: Looks good

PR #539, opened 2024-07-01
allow squads to override alerts about OOM to a specific channel

JensMagnus:


cmd/daemon/kubernetes/pods.go
```diff
@@ -267,5 +267,16 @@ func alertSquad(squad string, annotations map[string]string) (alertChannel strin
...
+	if destination, ok := squadCustomAlertChannel[squad]; ok {
+		return destination
+	}
+
 	return fmt.Sprintf("#squad-%s-alerts", squad)
 }
+
+var (
+	squadCustomAlertChannel = map[string]string{
+		"atlas": "squad-atlas-critical-alerts",
```
mahlunar, line 280: This should be a flag in the CLI so that is customisable when we define its deployment

mahlunar (commented):
Thanks for the contribution.

Be aware that the release manager is an open-source repo and its deployment is managed separately.

cmd/daemon/kubernetes/pods.go
JensMagnus: I've converted the PR back into draft, so let's adjust it to your requests as they make more sense. Thank you! :)

PR #551, opened 2024-11-11
Fix Grafana annotation formatting

JacobValdemar:
`\n` does not seem to create a line break in annotations in Grafana. `<br>` seems to create a line break when viewing the annotation. To verify this, try editing an annotation and adding a `<br>` in it. (Public repository, so not inserting a screenshot with proof)


kjuulh (approved):
Worth a shot xD

PR #556, opened 2025-02-28
feat: increase rollback artifact list length

Rosenberg96:
We want to see more artifacts to make it easier to select the correct artifact in rollbacks


mahlunar:
This is the upstream repository so we should make it configurable instead. Then we can choose the constant when deploying into our environmnet

Rosenberg96:
> This is the upstream repository so we should make it configurable instead. Then we can choose the constant when deploying into our environmnet

Fixed

mahlunar (approved):

PR #611, opened 2026-05-08
Adding slack release notifications for squads per env

mdurantejr:
- Adds Slack notification to squad release channels using Kubernetes squad labels, with the pattern `#squad-<squad>-releases-<env>`.
  - Release announcements now scan the manifests in the artifact and post to each distinct squad channel found.
  - Deployment success, pod error, and job error notifications now carry squad information from live Kubernetes labels and also post to the matching squad release channel.
  - The new squad-channel notification path is fully failsafe: if no squad label is present, the Slack channel does not exist, or Slack posting fails, the main flow continues unchanged and the error is only logged.
  - Existing notifications to `#releases-<env>`, direct user messages, and squad alert channels remain unchanged.

_Testing_

  - Added tests for manifest squad extraction and best-effort Slack notifications.
  - Ran go test ./...

Seel also: https://lunar.slack.com/archives/C04QA2ZUU1M/p1778138870000079

---

> [!NOTE]
> **Medium Risk**
> Adds new squad propagation across k8s event ingestion and Slack notification paths, including new best-effort fan-out behavior and new fields in shared event types. Main risk is unintended notification routing/noise or missed direct messages due to changed Slack error-handling paths.
> 
> **Overview**
> Adds *squad-aware* release and Kubernetes notifications by propagating a new `Squad` field through `http.ReleaseEvent`/`http.JobErrorEvent` and flow notify options, and populating it from Kubernetes `squad` labels (falling back to pod template labels).
> 
> Slack notifications now **fan out best-effort** to per-squad per-environment channels (`#squad-<squad>-releases-<env>`) for release announcements, deploy success, pod errors, and job errors; failures to post to squad channels are logged and do not fail the primary notification.
> 
> Release commits additionally attempt to derive `Squad` by scanning rendered manifest YAMLs in the artifact (`squadFromManifests`), with new tests covering manifest parsing/short-circuiting and Slack best-effort squad channel behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 96467d5108ca474a099282d9edd8404dfdb03261. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/daemon/kubernetes/kubernetes.go
+++ b/cmd/daemon/kubernetes/kubernetes.go
@@ -2,6 +2,7 @@ package kubernetes
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/lunarway/release-manager/internal/log"
 	"github.com/pkg/errors"
@@ -91,4 +92,21 @@ func observe(annotations map[string]string) {
 
 func isObserved(annotations map[string]string) bool {
 	return annotations[observedAnnotationKey] == annotations[artifactIDAnnotationKey]
+}
+
+func getSquadLabel(labels map[string]string) string {
+	if labels == nil {
+		return ""
+	}
+	return strings.TrimSpace(labels[squadLabelKey])
+}
+
+func firstNonEmpty(values ...string) string {
+	for _, value := range values {
+		value = strings.TrimSpace(value)
+		if value != "" {
+			return value
+		}
+	}
+	return ""
 }
--- a/cmd/daemon/kubernetes/kubernetes.go
+++ b/cmd/daemon/kubernetes/kubernetes.go
@@ -98,7 +98,10 @@ func getSquadLabel(labels map[string]string) string {
 	if labels == nil {
 		return ""
 	}
-	return strings.TrimSpace(labels[squadLabelKey])
+	if squad, ok := labels[squadLabelKey]; ok {
+		return strings.TrimSpace(squad)
+	}
+	return ""
 }
 
 func firstNonEmpty(values ...string) string {
--- a/cmd/daemon/kubernetes/pods.go
+++ b/cmd/daemon/kubernetes/pods.go
@@ -67,13 +67,15 @@ func (p *PodInformer) handle(obj interface{}) {
 	}
 
 	ctx := context.Background()
+	squad := getSquadLabel(pod.Labels)
+	alertSquadName := alertSquad(firstNonEmpty(squad, "no-one"), pod.Annotations)
 	event := http.PodErrorEvent{
 		PodName:     pod.Name,
 		Namespace:   pod.Namespace,
 		ArtifactID:  pod.Annotations[artifactIDAnnotationKey],
 		AuthorEmail: pod.Annotations[authorAnnotationKey],
-		Squad:       getCodeOwnerSquad(pod.Labels),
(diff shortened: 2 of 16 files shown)


cmd/daemon/kubernetes/kubernetes.go
```diff
@@ -92,3 +93,20 @@ func observe(annotations map[string]string) {
 func isObserved(annotations map[string]string) bool {
 	return annotations[observedAnnotationKey] == annotations[artifactIDAnnotationKey]
 }
+
+func getSquadLabel(labels map[string]string) string {
+	if labels == nil {
+		return ""
+	}
+	return strings.TrimSpace(labels[squadLabelKey])
```
mahlunar: I would prefer that we check if the map key exists so something like

```
if squad, ok := labels[squadLabelKey]; ok {
...
}
```

internal/flow/squads.go
```diff
@@ -0,0 +1,110 @@
...
+type manifestDocument struct {
+	Metadata manifestMetadata   `yaml:"metadata"`
+	Spec     manifestSpec       `yaml:"spec"`
+	Items    []manifestDocument `yaml:"items"`
+}
+
+func squadsFromManifests(dir string) ([]string, error) {
+	squadSet := make(map[string]struct{})
+
+	err := filepath.WalkDir(dir, func(path string, d fs.DirEntry, walkErr error) error {
```
mahlunar, line 42: This will possible walk through all files in the directory. What about adding an early return in case we find a squad?

internal/slack/slack.go
```diff
@@ -150,6 +151,7 @@ type ReleaseOptions struct {
 	CommitAuthor      string
 	CommitAuthorEmail string
 	Releaser          string
+	Squads            []string
```
mahlunar: I would not expect a scenario where one release can trigger notifications to multiple squads?

mdurantejr: This was initially written with the assumption that many squads can be listed, but I agree this may not be an usual case and have updated the code to handle a single squad/destination only.

internal/flow/squads.go
mdurantejr: This was for the case of multi-squads. Modified now to short-circuit on any match.

mdurantejr:
bugbot run

cmd/daemon/kubernetes/pods.go
```diff
@@ -67,13 +67,15 @@ func (p *PodInformer) handle(obj interface{}) {
...
+	alertSquadName := alertSquad(firstNonEmpty(squad, "no-one"), pod.Annotations)
 	event := http.PodErrorEvent{
 		PodName:     pod.Name,
 		Namespace:   pod.Namespace,
 		ArtifactID:  pod.Annotations[artifactIDAnnotationKey],
 		AuthorEmail: pod.Annotations[authorAnnotationKey],
-		Squad:       getCodeOwnerSquad(pod.Labels),
-		AlertSquad:  alertSquad(getCodeOwnerSquad(pod.Labels), pod.Annotations),
+		Squad:       squad,
+		AlertSquad:  alertSquadName,
```
tmablunar: Consider renaming to `SquadAlertChannel` or similar

tmablunar (approved):

PR #624, opened 2026-06-09
perf(git): use shallow clone for working-tree operations

mahlunar:
## Summary

Replace the full `cp -a` master repo copy (including `.git/` history) with a `git clone --depth 1 --local` for callers that only need to commit and push a single file. Saves ~1.9s on every release, conflict or not.

## Changes

- Add `ShallowClone(ctx, destination string) error` to `git.Service` — runs `git clone --depth 1 --local <masterPath> <destination>` then fixes the remote URL to `ConfigRepoURL` so subsequent pushes reach the real origin
- Add `ShallowClone` to `GitService` interface in both `flow` and `policy` packages
- Update 3 callers to use `ShallowClone`: `release.go` ×2, `policy.go`
- `describe_release.go` keeps using `Clone` since it walks the git log
- Remove now-unused `Clone` from `policy.GitService` interface and mock
- Update `branch_restriction_test.go` mock expectation accordingly

## Why

From `docs/trace.md`: `copyMaster` copies the full `.git/` history on every release (~1.9s). The per-release working dir only needs to commit one file and push — it never traverses history. This is fix #2 of 4 low-hanging-fruit items identified in the trace.

Closes DMAT-338

---

> [!NOTE]
> **Medium Risk**
> Touches release and policy commit/push paths; behavior should match prior flows but shallow clones differ from full copies if anything relied on full history in those temp dirs.
> 
> **Overview**
> Adds **`ShallowClone`** on `git.Service`: a depth-1 local clone from the in-memory master (`git clone --depth 1 --local`) followed by resetting `origin` to `ConfigRepoURL`, so commit/push still hit the real remote without copying full `.git` history.
> 
> **Release** and **policy** paths that only need a writable tree now call `ShallowClone` instead of full `Clone`/`copyMaster` (`release.go` for pre-check and `ExecReleaseArtifactID`, `policy.updatePolicies`). **`describe_release`** still uses **`Clone`** because it walks history. The **`policy.GitService`** interface drops `Clone` in favor of `ShallowClone`; flow keeps both. Mocks and `branch_restriction_test` expectations are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 61cca8abe6e7118d9ff1de29d9b52b329215e93c. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -139,17 +139,11 @@ func (s *Service) ShallowClone(ctx context.Context, destination string) error {
 	span, ctx := s.Tracer.FromCtx(ctx, "git.ShallowClone")
 	defer span.End()
 
-	span, _ = s.Tracer.FromCtx(ctx, "remove destination")
-	err := os.RemoveAll(destination)
-	span.End()
+	unlock, err := s.prepareDestination(ctx, destination)
 	if err != nil {
-		return errors.WithMessage(err, "remove existing destination")
+		return err
 	}
-
-	span, _ = s.Tracer.FromCtx(ctx, "lock mutex")
-	s.masterMutex.RLock()
-	defer s.masterMutex.RUnlock()
-	span.End()
+	defer unlock()
 
 	span, _ = s.Tracer.FromCtx(ctx, "shallow clone")
 	err = execCommand(ctx, filepath.Dir(destination), "git", "clone", "--depth", "1", "--local", s.masterPath, destination)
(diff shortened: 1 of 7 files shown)


internal/git/git.go
```diff
@@ -131,6 +132,42 @@ func (s *Service) Clone(ctx context.Context, destination string) (*git.Repositor
...
+	span.End()
+	if err != nil {
+		return errors.WithMessage(err, "remove existing destination")
+	}
+
+	span, _ = s.Tracer.FromCtx(ctx, "lock mutex")
+	s.masterMutex.RLock()
+	defer s.masterMutex.RUnlock()
+	span.End()
+
```
tmablunar, line 147: some of this could perhaps be DRY'ed up with `copyMaster`

tmablunar (approved):

PR #637, opened 2026-06-22
perf(flow): drop redundant SyncMaster in retry handler

mahlunar:
## What

Removes the redundant `SyncMaster` call in `flow.Service.retry`'s push-conflict handler. The retry wrapper now delegates straight to `try.Do`.

## Why

On a push conflict, `Commit` → `rebaseOntoMaster` (`git.go:444`) **already** resyncs the local master mirror (`SyncMaster`, ~6s under the write lock) before surfacing `ErrBranchBehindOrigin`. The retry handler then fired a **second** `SyncMaster` — redundant in the common case, since `rebaseOntoMaster` just synced. That second call held the write lock for another ~2–3.5s per conflict, and Go's `RWMutex` blocks new readers once a writer queues, so it starved `ShallowClone` across all parallel workers.

Removing it lets the next retry's `ShallowClone` proceed without that extra write-lock hold.

**Expected impact:** saves ~2–3.5s per push conflict. This is a contention fix, so the gain scales with how often workers actually collide on `master` — it does nothing when there's no conflict. It's cheap, safe, and negative-LOC, so worth shipping regardless.

## How it's safe

The retried function only surfaces `ErrBranchBehindOrigin` through `s.Git.Commit` (`release.go:233`), and `Commit` only returns that error after going through `rebaseOntoMaster` (`git.go:400`), which calls `SyncMaster` first. There are two such return sites, both downstream of the rebase:

- `git.go:402` — the rebase (or the `SyncMaster` inside it) failed.
- `git.go:406` — the push *after* a successful rebase was itself rejected because origin advanced again.

So the mirror is **not** guaranteed fresh on the next attempt: origin can advance after the inner `SyncMaster` (the line 406 case is positive evidence of exactly that), or the inner `SyncMaster` can itself fail and leave the mirror unsynced. The safety doesn't rely on freshness — it relies on **self-healing**: whatever the mirror's state, the next retry runs the full clone → commit → push again; if the mirror is stale the push is rejected, which re-enters `rebaseOntoMaster` → `SyncMaster`. The removed handler-level `SyncMaster` was therefore redundant, costing at most one extra retry iteration in rare interleavings — never incorrect behavior or a stuck loop.

Part of DMAT-421

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Changes release retry behavior around git push conflicts; incorrect assumption that the mirror is always fresh could cause extra failed retries, though the PR argues the only `ErrBranchBehindOrigin` path already syncs in `git.Commit`.
> 
> **Overview**
> **Removes a second `SyncMaster` on push conflicts** in `flow.Service.retry`, which was holding the master write lock an extra ~2–3.5s and starving parallel `ShallowClone` readers.
> 
> `retry` now passes through to `try.Do` only. The special case that detected `ErrBranchBehindOrigin`, logged, and called `Git.SyncMaster` before retrying is gone, along with the `internal/git` import in `flow.go`. A comment documents that **`Commit` → `rebaseOntoMaster` already refreshes the mirror** before that error reaches the retry loop, so the next attempt’s `ShallowClone` can use the updated mirror without another sync.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6f1fd1d0d8ea88ca202ab7267e0b8f2643a1ec07. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/flow/flow.go
+++ b/internal/flow/flow.go
@@ -19,7 +19,6 @@ import (
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/lunarway/release-manager/internal/artifact"
 	"github.com/lunarway/release-manager/internal/copy"
-	internalgit "github.com/lunarway/release-manager/internal/git"
 	httpinternal "github.com/lunarway/release-manager/internal/http"
 	"github.com/lunarway/release-manager/internal/intent"
 	"github.com/lunarway/release-manager/internal/log"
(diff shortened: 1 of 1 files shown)


internal/flow/flow.go
```diff
@@ -123,23 +122,15 @@ type GitService interface {
...
-				log.WithContext(ctx).Infof("flow/retry: master repo not aligned with origin. Syncing and retrying")
-				err := s.Git.SyncMaster(ctx)
-				if err != nil {
-					return false, errors.WithMessage(err, "sync master")
-				}
-			}
-		}
-		return stop, err
-	})
+	return try.Do(ctx, s.Tracer, s.MaxRetries, f)
```
tmablunar, line 133: What can change between retries? Is there any reason to retry or should it just fail on error?

tmablunar (approved):

internal/flow/flow.go
mahlunar: We still need to retry. If the branch is behind then it is fast-forwarded already so retrying would most likely succeed the next time.

PR #633, opened 2026-06-19
perf(amqp): configurable consumer workers

mahlunar:
## Summary

Increases AMQP consumer parallelism from 1 worker to 5, allowing release flows to be processed concurrently and reducing queue latency under burst load.

## Changes

- `WorkerCount`: 1 → 5 (hardcoded), now threaded through `Config.WorkerCount`
- `Prefetch`: 0 → configurable count, now threaded through `Config.Prefetch`
- Adds `--amqp-worker-count` CLI flag (default: 1) to `cmd/server/command/root.go`
- Updates `--amqp-prefetch` default from 1 → 5 to match worker count
- Wires both values through `amqpOptions` → `amqpextra.Config` → `StartConsumer` so operators can tune without a code change

## Why

Each flow gets its own isolated temp directory — workers don't share working trees. The master git mirror uses `sync.RWMutex`; `ShallowClone` takes an `RLock` so all workers clone concurrently without blocking each other. Push conflicts are already handled via rebase-on-conflict retry.

The previous `Prefetch: 0` (unlimited) would have overwhelmed workers under burst load; capping prefetch to equal WorkerCount ensures the broker only buffers as many messages as workers can process.

Closes DMAT-415

---

> [!NOTE]
> **Medium Risk**
> Raises concurrent release/git work on the shared config repo; isolation and retry logic mitigate but push races and load scale with worker count.
> 
> **Overview**
> **AMQP release processing can run with up to five concurrent consumer workers** instead of a single hardcoded worker, with defaults and flags so operators can tune parallelism without code changes.
> 
> New **`--amqp-worker-count`** (default **5**) is wired from server CLI → `amqpOptions` → `amqpextra.Config` → `internalamqp.ConsumerConfig.WorkerCount`. **`--amqp-prefetch`** default moves **1 → 5**, and the consumer now uses **`Config.Prefetch`** instead of a fixed **0** (unlimited), so prefetch is expected to match worker count and cap in-flight messages under burst load.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e0bab58af31f1458f2a35a98f2aece1360d6772b. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #638, opened 2026-06-22
perf(git): serialize push step with mutex to eliminate branch conflicts

mahlunar:
## Summary

Adds a size-1 channel mutex to `git.Service` that serializes `git push origin` calls across `Commit` and `SignedCommit`. This eliminates branch-behind-origin conflicts between parallel workers without reducing clone/commit parallelism.

## Why

After increasing AMQP consumers from 1 to 5 parallel workers (#633), release flow p99 latency jumped to ~30s. Root cause: 5 workers pushing to GitHub concurrently meant workers 2–5 got rejected with "branch behind origin", triggering `rebaseOntoMaster` → `SyncMaster` write-lock cascades (~8–10s each). Serializing only the push step (which is the actual race point) eliminates the conflicts entirely; the expensive clone (~1.5s) and commit steps remain parallel.

Closes DMAT-421

---

> [!NOTE]
> **Medium Risk**
> Touches the release git push path used by multiple workers, but the change is narrowly scoped serialization with the existing rebase-on-conflict fallback unchanged.
> 
> **Overview**
> Adds a **lazy-initialized size-1 `pushSem`** on `git.Service` so concurrent workers no longer race on `git push origin` at the same time.
> 
> `Commit` and `SignedCommit` now **acquire/release the semaphore around every `gitPush` call**, including the post-rebase retry in `Commit`, while clone and commit work stay parallel. A unit test checks that only one goroutine holds the semaphore at once.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d3096c6f9cf51b750fac102a63673ffc6aa67e44. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -46,10 +46,20 @@ type Service struct {
 	masterPath  string
 	masterMutex sync.RWMutex
 	master      *git.Repository
+
+	// pushSem serializes git push origin calls to prevent branch-behind-origin
+	// conflicts when multiple workers push concurrently.
+	pushSem     chan struct{}
+	pushSemOnce sync.Once
 }
 
 func (s *Service) MasterPath() string {
 	return s.masterPath
+}
+
+func (s *Service) getPushSem() chan struct{} {
+	s.pushSemOnce.Do(func() { s.pushSem = make(chan struct{}, 1) })
+	return s.pushSem
 }
 
 // InitMasterRepo clones the configuration repository into a master directory.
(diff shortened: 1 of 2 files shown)


internal/git/git.go
```diff
@@ -46,12 +46,22 @@ type Service struct {
 	masterPath  string
 	masterMutex sync.RWMutex
 	master      *git.Repository
+
+	// pushSem serializes git push origin calls to prevent branch-behind-origin
+	// conflicts when multiple workers push concurrently.
+	pushSem     chan struct{}
```
mahlunar: should be sync.Mutex instead

tmablunar (approved):
Description still mentions channel

